PR #4, opened 2018-02-26
Update readme and license and add contributing

rossf7:


LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2016 GiantSwarm GmbH
```
fgimenez: shouldn't this be 2018?

rossf7: Good catch I was in a time warp! Fixed.

rossf7:
There is `repo-init` but it's very old and is Ruby unlike most of our tools. 

https://github.com/giantswarm/repo-init

It has some useful features like setting up repo labels. An updated tool would be good. `operatorkit init` would also be nice. As there is quite a lot of plumbing needed to set up an operator.

fgimenez:
You are totally right, most of the files in this PR should belong to `repo-init` :+1:

PR #7, opened 2018-03-06
Use latest operatorkit interface

rossf7:
Updating to the latest interface before we start implementing the chart resource.


rossf7:
Towards giantswarm/giantswarm/issues/2747

PR #12, opened 2018-03-07
add GetRelease method

fgimenez:

Adds initial client code and `GetRelease` method required by the desired state calculations in the chart resource.

Looking at CNR API it's very easy to write a REST client, I've taken this approach here instead of using helm plus the registry plugin (or the appr python code directly). 

I'll propose next the wiring with the chart resource, we need an interface in order to setup proper mocking and not having to spin up a server for chart unit tests. We'll have soon in place e2e tests with an actual CNR server from https://github.com/giantswarm/docker-cnr-server.

### DIFF
--- a/service/chartconfig/v1/appr/appr.go
+++ b/service/chartconfig/v1/appr/appr.go
@@ -44,7 +44,7 @@ func New(config Config) (*Client, error) {
 
 	// set client timeout to prevent leakages.
 	hc := &http.Client{
-		Timeout: time.Second * 5,
+		Timeout: time.Second * httpClientTimeout,
 	}
 
 	u, err := url.Parse(config.Address + "/" + config.Organization + "/")
--- a/service/chartconfig/v1/appr/appr.go
+++ b/service/chartconfig/v1/appr/appr.go
@@ -2,6 +2,14 @@
 package appr
 
 import (
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
+	"github.com/giantswarm/chart-operator/service/chartconfig/v1/key"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 )
@@ -10,14 +18,16 @@ import (
 type Config struct {
 	Logger micrologger.Logger
 
-	Address string
+	Address      string
+	Organization string
 }
 
 // Client knows how to talk with a CNR server.
 type Client struct {
-	logger micrologger.Logger
+	httpClient *http.Client
+	logger     micrologger.Logger
 
-	address string
+	base *url.URL
 }
 
 // New creates a new configured appr client.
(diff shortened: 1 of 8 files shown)


service/chartconfig/v1/appr/appr_test.go
```diff
@@ -0,0 +1,99 @@
...
+func TestGetRelease(t *testing.T) {
+	customObject := v1alpha1.ChartConfig{
+		Spec: v1alpha1.ChartConfigSpec{
+			Chart: v1alpha1.ChartConfigSpecChart{
+				Name: "chartname",
+			},
+		},
+	}
+	tcs := []struct {
+		description     string
```
MarcelMue: Would be nice to align these with the (not yet merged) test patterns from fmt. 
https://github.com/giantswarm/fmt/pull/5

service/chartconfig/v1/appr/appr.go
```diff
@@ -10,31 +18,97 @@ import (
...
+	if config.Organization == "" {
+		return nil, microerror.Maskf(invalidConfigError, "organization must not be empty")
+	}
+
+	// set client timeout to prevent leakages.
+	hc := &http.Client{
+		Timeout: time.Second * 5,
+	}
+
+	u, err := url.Parse(config.Address + "/" + config.Organization + "/")
```
MarcelMue: I agree that this is easy to do instead of using helm + the plugin. My only concern would be that an update of the plugin where this URL is different, would now break.
Nothing major, so I am fine with this.

service/chartconfig/v1/appr/spec.go
```diff
@@ -0,0 +1,5 @@
+package appr
+
+type Package struct {
```
indyfree: Just for personal interest: Why is this named `Package`? I've only seen `Application` in quay.io so far, is this CNR terminology?

service/chartconfig/v1/appr/appr.go
fgimenez: :+1: we need to keep an eye on the CNR specification and adapt to it as needed. Using the plugin we are exposed to the same problem though. I hope that any eventual change will be backward compatible.

service/chartconfig/v1/appr/appr_test.go
fgimenez: I think this test mostly adhere to the proposed patterns. For this specific field and this specific test IMO `description` matches better the intention, these test cases are not trying to verify that the code works well, they served as a guide during development and now are documenting the expected behavior of the code. In this sense, each test case `describes` one of the facets of the code, WDYT?

kopiczko: I'm with @MarcelMue if you want to change `name` for `description` that's ok. Just propose it in fmt and let's use that everywhere.

MarcelMue: I also really like the `switch case` for error checking, this is my preference though - I am not going to block your PR over it.

service/chartconfig/v1/appr/spec.go
fgimenez: Exactly, `Package` is a concept from CNR, see for instance https://github.com/app-registry/appr/blob/master/appr/api/impl/registry.py#L198

service/chartconfig/v1/appr/appr.go
```diff
@@ -10,31 +18,97 @@ import (
...
 	if config.Address == "" {
 		return nil, microerror.Maskf(invalidConfigError, "address must not be empty")
 	}
+	if config.Organization == "" {
+		return nil, microerror.Maskf(invalidConfigError, "organization must not be empty")
+	}
+
+	// set client timeout to prevent leakages.
+	hc := &http.Client{
+		Timeout: time.Second * 5,
```
rossf7: Does it make sense to have a constant or a flag for the timeout?

service/chartconfig/v1/appr/appr_test.go
fgimenez: Ok, didn't know about the `switch`, thx! added. 

I still think that the field name should not be imposed, in this case it makes more sense to me to use `description` but in other cases might be better `name` or even having more than one documenting fields (`given`, `when` and `then` would be very nice), it depends on the testing style.

service/chartconfig/v1/appr/appr.go
fgimenez: Totally, added a constant for now.

PR #13, opened 2018-03-08
define appr interface and wire with chart resource

fgimenez:

Connects appr service and chart resource. I've added also an interface to appr in order to have both components properly decoupled.

### DIFF
--- a/service/chartconfig/v1/appr/appr.go
+++ b/service/chartconfig/v1/appr/appr.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"path"
 	"time"
 
 	"github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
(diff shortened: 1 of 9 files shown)


rossf7:
Actually the mock is already there :) Nice!

fgimenez:
> For me, the previous name makes more sense because we are getting the release that has been pushed to the specified channel.

Thanks a lot for pointing it out, previously the client was getting information about the package without including channel info, that's why the returned release was the default one, which led me to rename the method. I've changed it back the `GetRelease` and added channel info to the query, PTAL.

service/chartconfig/v1/appr/appr.go
```diff
@@ -63,23 +64,26 @@ func New(config Config) (*Client, error) {
 }
 
 // GetRelease queries CNR for the release name of the chart represented by the
-// given custom object
+// given custom object (including channel info)
```
rossf7: Nitpick, needs a `.`.

fgimenez: Sure, done.

PR #14, opened 2018-03-09
Use latest operatorkit

rossf7:
Vendor update and adapt to recent changes.


rossf7:
Jeje :) I was in there already and wanted to get us up to date.

PR #18, opened 2018-03-14
Add Helm service and get Release status

rossf7:

This PR adds the Helm service which uses the Helm Go client to get the Release status. We need this to know if the Chart has been deployed.


service/chartconfig/v1/helm/helm.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return newHelm, nil
+}
+
+// GetReleaseContent gets the current status of the Helm Release.
+func (c *Client) GetReleaseContent(customObject v1alpha1.ChartConfig) (*Release, error) {
+	releaseName := key.ReleaseName(customObject)
+
+	resp, err := c.helmClient.ReleaseContent(releaseName)
+	if strings.HasPrefix(err.Error(), notFoundErrorPrefix) {
```
xh3b4sd: I would like to deal with the error handling magic like this: https://github.com/giantswarm/e2e-app/blob/c977516cbc3bc4fe7af598cc3841d884d136930c/error.go#L9-L24. Then we stick to our concept of a private error type and a public error matcher, which can check however it likes. Then the conditionals here are cleaner.

service/chartconfig/v1/helm/helm.go
```diff
@@ -0,0 +1,71 @@
...
+		return nil, microerror.Maskf(notFoundError, "Release: %s not found", releaseName)
```
xh3b4sd: IMO it is enough to annotate the error with the release name. It then reads something like `not found: release-name`. When we want to be more explicit I would like to adapt the error itself to `releaseNotFoundError`, which then reads like `release not found: release-name`. Now the error would be rendered like `not found: Release: release-name not found`, which stutters and is a bit messy.

service/chartconfig/v1/helm/helm.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	release := &Release{
+		Name:   resp.Release.Name,
+		Status: resp.Release.Info.Status.Code.String(),
```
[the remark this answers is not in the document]
xh3b4sd: For my interest, what is that status telling us?

service/chartconfig/v1/helm/helm.go
rossf7: 👍  I'll remove the stuttering.

service/chartconfig/v1/helm/helm.go
rossf7: Nice, that's cleaner I'll do that.

service/chartconfig/v1/helm/helm.go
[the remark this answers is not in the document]
rossf7: It's the status of the Helm Release. We need it to know whether we should install or update the chart.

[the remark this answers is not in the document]
xh3b4sd: I think I was looking for this. 

> state of the release (can be: UNKNOWN, DEPLOYED, DELETED, SUPERSEDED, FAILED or DELETING)

See https://github.com/kubernetes/helm/blob/b335bfdabf03cc6aaea9bd3e46a5fc63c932c1dd/docs/helm/helm_status.md.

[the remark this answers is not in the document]
rossf7: Yes that's it.

service/chartconfig/v1/helm/spec.go
```diff
@@ -0,0 +1,16 @@
+package helm
+
+import "github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
+
+// Interface describes the methods provided by the helm client.
+type Interface interface {
```
fgimenez: Nice!

service/chartconfig/v1/helm/helm.go
```diff
@@ -0,0 +1,71 @@
...
+func New(config Config) (*Client, error) {
+	if config.Logger == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
+	}
+	if config.Host == "" {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Host must not be empty", config)
+	}
+
+	hc := helmclient.NewClient(helmclient.Host(config.Host),
+		helmclient.ConnectTimeout(connectionTimeoutSecs))
```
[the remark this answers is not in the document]
fgimenez: Awesome! So easy to import, instantiate and, below, access the exposed methods <3 helm

main.go
```diff
@@ -137,10 +137,11 @@ func mainWithError() (err error) {
 
 	daemonCommand := newCommand.DaemonCommand().CobraCommand()
 
-	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.Address, "", "Address used to connect to Kubernetes. When empty in-cluster config is created.")
 	daemonCommand.PersistentFlags().String(f.Service.CNR.Address, "https://quay.io", "Address used to connect to CNR, defaults to quay's managed offering.")
 	daemonCommand.PersistentFlags().String(f.Service.CNR.Organization, "giantswarm", "CNR organization.")
-	daemonCommand.PersistentFlags().Bool(f.Service.Kubernetes.InCluster, true, "Whether to use the in-cluster config to authenticate with Kubernetes.")
+	daemonCommand.PersistentFlags().String(f.Service.Helm.Host, ":44134", "Host and port for Helm client to connect to the Tiller endpoint.")
```
fgimenez: Would be good to have an integration test for checking the connection, in the current tests tiller is already set up available.

rossf7: Definitely! I'd like to tackle that after I've got GetCurrentState working. I'd like to get my feet wet with integration testing :)

service/chartconfig/v1/helm/helm_test.go
```diff
@@ -0,0 +1,60 @@
...
+				},
+			},
+			expectedRelease: nil,
+			errorMatcher:    IsNotFound,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			helm := Client{
```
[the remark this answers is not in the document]
rossf7: Done.

service/chartconfig/v1/helm/helm.go
rossf7: Done.

rossf7:
Should be good now. I like the error handling much more now ❤️ .

@xh3b4sd PTAL

service/chartconfig/v1/helm/error.go
```diff
@@ -9,9 +17,16 @@ func IsInvalidConfig(err error) bool {
 	return microerror.Cause(err) == invalidConfigError
 }
 
-var notFoundError = microerror.New("not found")
+var releaseNotFoundError = microerror.New("release not found")
+
+// IsReleaseNotFound asserts releaseNotFoundError.
+func IsReleaseNotFound(err error) bool {
+	if strings.HasPrefix(err.Error(), releaseNotFoundErrorPrefix) {
```
xh3b4sd: When working with errors within matchers it is always safer to only work with the cause like below. The error here can be wrapped and the message might not be what we expect because we do not match against the origin of the error. This is actually the task of this matcher though.

service/chartconfig/v1/helm/helm.go
```diff
@@ -55,8 +51,8 @@ func (c *Client) GetReleaseContent(customObject v1alpha1.ChartConfig) (*Release,
 	releaseName := key.ReleaseName(customObject)
 
 	resp, err := c.helmClient.ReleaseContent(releaseName)
-	if strings.HasPrefix(err.Error(), notFoundErrorPrefix) {
-		return nil, microerror.Maskf(notFoundError, "Release: %s not found", releaseName)
+	if IsReleaseNotFound(err) {
+		return nil, microerror.Maskf(releaseNotFoundError, "%s", releaseName)
```
xh3b4sd: We can just put the release name variable. No need for the format string here unless `releaseName` is not already a string.

rossf7:
Error handling updated to use Cause and fix the formatting. CI is green going with this.

PR #22, opened 2018-03-15
Get release content

rossf7:

Updates `GetReleaseContent` to get the status and any values that have been set when the Chart was installed.

### DIFF
--- a/service/chartconfig/v1/helm/helm.go
+++ b/service/chartconfig/v1/helm/helm.go
@@ -4,6 +4,7 @@ import (
 	"github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
+	"k8s.io/helm/pkg/chartutil"
 	helmclient "k8s.io/helm/pkg/helm"
 
 	"github.com/giantswarm/chart-operator/service/chartconfig/v1/key"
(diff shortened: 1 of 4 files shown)


service/chartconfig/v1/helm/helm.go
```diff
@@ -46,21 +47,30 @@ func New(config Config) (*Client, error) {
...
 		return nil, microerror.Mask(err)
 	}
 
+	// Raw values are returned by the Tiller API and we convert these to a map.
+	raw := []byte(resp.Release.Config.Raw)
+	values, err := chartutil.ReadValues(raw)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
```
rossf7: We need this foo to get the values. We use the Helm `chartutil` package rather than doing it ourselves.

service/chartconfig/v1/helm/helm.go
```diff
@@ -46,21 +47,30 @@ func New(config Config) (*Client, error) {
...
 	release := &Release{
 		Name:   resp.Release.Name,
 		Status: resp.Release.Info.Status.Code.String(),
+		Values: values.AsMap(),
```
rossf7: `AsMap` returns the structure we need.

service/chartconfig/v1/helm/helm_test.go
```diff
@@ -7,17 +7,73 @@ import (
...
+					Name:      "chart-operator",
+					Namespace: "default",
+				}),
+			},
+			expectedRelease: &Release{
+				Name:   "chart-operator",
+				Status: "DEPLOYED",
+				Values: map[string]interface{}{
+					// Note: Values cannot be configured via the Helm mock client.
+					"name": "value",
```
rossf7: A shame the Helm mock doesn't cover this.

service/chartconfig/v1/helm/helm_test.go
```diff
@@ -7,17 +7,73 @@ import (
...
+						Channel: "0.1-beta",
+						Release: "chart-operator",
+					},
+				},
+			},
+			releases: []*helmrelease.Release{
+				helmclient.ReleaseMock(&helmclient.MockReleaseOptions{
+					Name:       "chart-operator",
+					Namespace:  "default",
+					StatusCode: helmrelease.Status_FAILED,
```
rossf7: The Helm client has an enumeration for the status which I like. Not certain yet but I think our Helm client package could also expose this to the `chart` resource.

service/chartconfig/v1/resource/chart/resource_test.go
```diff
@@ -16,20 +16,23 @@ func Test_toChartState(t *testing.T) {
...
 			},
 		},
 		{
 			name: "case 1: wrong type",
 			input: ChartState{
 				ChartName:      "test-chart",
 				ChannelName:    "test-channel",
-				ReleaseVersion: "test-release",
+				ReleaseName:    "test-release",
+				ReleaseVersion: "0.1.0",
```
rossf7: Just some tidying up. We added `ReleaseName` to `ChartState` but didn't update the tests.

service/chartconfig/v1/helm/helm.go
indyfree: Just for clearity: `values` are the values provided by the helm client overriding the charts default `values.yaml`?

rossf7: Yes it just returns the set values which is good for our use-case. We'll need the values in the CRD and will reconcile them to update the installed chart when they change.

PR #37, opened 2018-03-28
Add DeleteRelease to helm client

indyfree:

Add method for uninstalling a charts to the helm client.

Next steps would be to implement the corresponding chart resource methods.
```
func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange interface{}) 
func (r *Resource) NewDeletePatch(ctx context.Context, obj, currentState, desiredState 
```

### DIFF
--- a/service/chartconfig/v1/helm/helm.go
+++ b/service/chartconfig/v1/helm/helm.go
@@ -44,6 +44,16 @@ func New(config Config) (*Client, error) {
 	return newHelm, nil
 }
 
+// DeleteRelease uninstalls a chart given its release name
+func (c *Client) DeleteRelease(releaseName string, options ...helmclient.DeleteOption) error {
+	_, err := c.helmClient.DeleteRelease(releaseName, options...)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	return nil
+}
+
 // GetReleaseContent gets the current status of the Helm Release including any
 // values provided when the chart was installed. The releaseName is the name
 // of the Helm Release that is set when the Helm Chart is installed.
--- a/service/chartconfig/v1/helm/helm.go
+++ b/service/chartconfig/v1/helm/helm.go
@@ -44,7 +44,7 @@ func New(config Config) (*Client, error) {
 	return newHelm, nil
 }
 
-// DeleteRelease uninstalls a chart given its release name
+// DeleteRelease uninstalls a chart given its release name.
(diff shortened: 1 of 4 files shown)


service/chartconfig/v1/helm/helm.go
```diff
@@ -44,6 +44,16 @@ func New(config Config) (*Client, error) {
 	return newHelm, nil
 }
 
+// DeleteRelease uninstalls a chart given its release name
+func (c *Client) DeleteRelease(releaseName string, options ...helmclient.DeleteOption) error {
```
indyfree: Thats the exact name as in the helm package. Is that fine? Cause the other functions have different names than in the package.

service/chartconfig/v1/helm/helm_test.go
```diff
@@ -10,6 +10,67 @@ import (
...
+			releaseName: "chart-operator",
+			releases: []*helmrelease.Release{
+				helmclient.ReleaseMock(&helmclient.MockReleaseOptions{
+					Name:      "chart-operator",
+					Namespace: "default",
+				}),
+			},
+		},
+		{
+			description: "case 2: try to delete release with wrong name",
```
indyfree: Is that over the top? Technically its like `case 0`, just that theres another release installed.

indyfree:
Oh just see, build is failing. Nevermind then. Will first fix that

indyfree:
Alright, fixed the build :+1:

service/chartconfig/v1/helm/helm.go
rossf7: I think using the same name is fine. We do that for ReleaseStatus and ReleaseContent. InstallTarball is different because it uses the tarball we download from CNR.

service/chartconfig/v1/helm/helm_test.go
rossf7: No I like it. It checks we're deleting the correct release and not just the first one for example.

service/chartconfig/v1/helm/spec.go
```diff
@@ -4,6 +4,8 @@ import "k8s.io/helm/pkg/helm"
 
 // Interface describes the methods provided by the helm client.
 type Interface interface {
+	// DeleteRelease uninstalls a chart given its release name
```
rossf7: Nitpick `.` here please.

indyfree:
merged #37.

PR #43, opened 2018-04-06
Add chartconfig resource chart

fgimenez:


fgimenez:
Yep, TIL you can have multiple charts but their names must match `<project-name>-*-chart`, fixed.

PR #45, opened 2018-04-09
Make watch namespace configurable

rossf7:

Since we are going to run chart-operator in guest clusters we are going to install it in kube-system. The chartconfig CRs should also be in kube-system rather than the default namespace.

This PR makes the watch namespace configurable.


service/chartconfig/chart_framework.go
```diff
@@ -42,6 +43,9 @@ func NewChartFramework(config ChartFrameworkConfig) (*framework.Framework, error
 	if config.Logger == nil {
 		return nil, microerror.Maskf(invalidConfigError, "config.Logger must not be empty")
 	}
+	if config.ProjectName == "" {
+		return nil, microerror.Maskf(invalidConfigError, "config.ProjectName must not be empty")
+	}
```
rossf7: This check was missing. There is no check for WatchNamespace because "" is a valid value.

rossf7:
Ack. Blocks are separated now.

PR #56, opened 2018-04-23
Use helm and appr clients in e2e tests

indyfree:


integration/test/basic/basic_test.go
```diff
@@ -48,33 +50,40 @@ func TestChartInstalled(t *testing.T) {
 	}
 }
 
-func installChartOperatorResource(f *framework.Host) error {
+func installChartOperatorResource(f *framework.Host, helmClient *helmclient.Client) error {
 	const chartOperatorResourceValues = `chart:
   name: "tb-chart"
   channel: "5-5-beta"
   namespace: "default"
   release: "tb-release"
```
indyfree: I tried this notation: https://github.com/kubernetes/helm/blob/077be881c4ccb0e89e33432a286f20f47b9af17f/pkg/helm/helm_test.go#L110

But it didn't work (just wouldn't find release with no error) and I couldn't determine why. Anybody an idea?

PR #58, opened 2018-04-25
Add app and service-type labels

rossf7:
Adds the service-type label. There were also a few missing app labels.

### DIFF
--- a/helm/chart-operator-chart/Chart.yaml
+++ b/helm/chart-operator-chart/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for the chart-operator
 name: chart-operator-chart
-version: 0.1.0
+version: 0.1.2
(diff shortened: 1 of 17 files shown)


helm/chart-operator-chart/templates/deployment.yaml
```diff
@@ -14,8 +15,9 @@ spec:
     metadata:
       labels:
         app: {{ .Values.name }}
-        annotations:
-          checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        giantswarm.io/service-type: "managed"
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
```
rossf7: I messed up adding the annotation which broke CI. Fixed here and its now green.

helm/chart-operator-chart/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for the chart-operator
 name: chart-operator-chart
-version: 0.1.0
+version: 0.1.2
```
rossf7: The version needed to be bumped.

PR #60, opened 2018-04-26
Use correct RESTClient in controller config

MarcelMue:
This most likely caused Integration tests to fail, no clue how they could have ever worked TBH.

Might be worth a try to check if we reuse envs between tests which caused this flakyness.

Explanation can be found here: https://github.com/giantswarm/operatorkit/blob/1ddbbec19f4b9a9f8f942cb9313fc273e2a6a63c/controller/controller.go#L47

### DIFF
--- a/service/controller/chart.go
+++ b/service/controller/chart.go
@@ -97,7 +97,7 @@ func NewChart(config ChartConfig) (*Chart, error) {
 			Informer:       newInformer,
 			Logger:         config.Logger,
 			ResourceRouter: resourceRouter,
-			RESTClient:     config.K8sClient.CoreV1().RESTClient(),
+			RESTClient:     config.G8sClient.CoreV1alpha1().RESTClient(),
 
 			Name: config.ProjectName + chartControllerSuffix,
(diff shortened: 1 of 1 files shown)


MarcelMue:
Build is green now, but there is still some unrelated flapping /cc @fgimenez

fgimenez:
Great, thanks a lot!! Where do you see this flapping?

MarcelMue:
It just sometimes seems to time out with 
```
--- FAIL: TestChartInstalled (918.58s)
    basic_test.go:44: expected nil found could not retrieve release content: tb-release: release not found
```
But I have no idea about how this release magic works

fgimenez:
Do you have a link to a failing build? We should not have flaky tests at all, if a test sometimes fails we should disable it until we know what's going on, otherwise the tests don't give us any value.

MarcelMue:
Here you go: https://circleci.com/gh/giantswarm/chart-operator/476

fgimenez:
Thx, will try to reproduce locally.

PR #62, opened 2018-05-02
Add e2e test for chart deletion

indyfree:

Add deletion of a chart to a lifecycle test. Also provides a fix for (before failing) test in the resource deletion method.


indyfree:
Closing: Will Split up PR into fix and test.

PR #63, opened 2018-05-02
Add e2e deletion test

indyfree:

Add Deletion to a `Lifecycle` test, since we need to create the chart beforehand anyways.

Commit history looks weird because of merging and reverting with parallel branch to fix deletion errros.


integration/test/basic/basic_test.go
```diff
@@ -14,35 +14,53 @@ import (
...
 
-	var rc *helmclient.ReleaseContent
+	err = waitForReleaseStatus(helmClient, testRelease, "DELETED")
+	if !helmclient.IsReleaseNotFound(err) {
+		t.Fatalf("%q not succesfully deleted %v", testRelease, err)
+	}
+	log.Printf("%q succesfully deleted", testRelease)
+}
+
+func waitForReleaseStatus(helmClient *helmclient.Client, release string, status string) error {
```
indyfree: Refactor to prepare for Upgrade Test where we need to match a `Update` Status.

integration/test/basic/basic_test.go
```diff
@@ -14,35 +14,53 @@ import (
...
+	log.Printf("deleting %q", cr)
+	err = helmClient.DeleteRelease(cr)
 	if err != nil {
-		t.Fatalf("could not install chart-operator-resource-chart %v", err)
+		t.Fatalf("could not delete %q %v", cr, err)
 	}
 
-	var rc *helmclient.ReleaseContent
+	err = waitForReleaseStatus(helmClient, testRelease, "DELETED")
+	if !helmclient.IsReleaseNotFound(err) {
```
indyfree: Release gets purged so in fact we are don't expect a "DELETED" Status, but expecting a error. Deletion takes a bit so we use the `waitForReleaseStatus` to wait for the error

integration/test/basic/basic_test.go
```diff
@@ -14,35 +14,53 @@ import (
...
-		log.Printf("waiting for release %s: %v", t, err)
+		log.Printf("getting release status %s: %v", t, err)
 	}
 
-	err = backoff.RetryNotify(operation, backoff.NewExponentialBackOff(), notify)
-	if err != nil {
-		t.Fatal("expected nil found", err)
-	}
+	eb := backoff.NewExponentialBackOff()
+	eb.MaxElapsedTime = time.Minute
```
indyfree: Don't wait for the default `15` minutes, since we need to match the error, and 1 minute has been enough so far. We can iteratate on that.

integration/test/basic/basic_test.go
```diff
@@ -27,39 +32,55 @@ func TestChartInstalled(t *testing.T) {
...
+	if !helmclient.IsReleaseNotFound(err) {
+		t.Fatalf("%q not succesfully deleted %v", testRelease, err)
+	}
+	log.Printf("%q succesfully deleted", testRelease)
+}
+
+func waitForReleaseStatus(gsHelmClient *helmclient.Client, release string, status string) error {
 	operation := func() error {
-		rc, err = helmClient.GetReleaseContent("tb-release")
+		rc, err := gsHelmClient.GetReleaseContent(release)
```
indyfree: Refactor Status checks to prepare for `Update` test. Note that the `gsHelmClient` is used here since guest services will be installed with our own tiller.

integration/test/basic/basic_test.go
```diff
@@ -27,39 +32,55 @@ func TestChartInstalled(t *testing.T) {
...
+	err = helmClient.DeleteRelease(cr)
 	if err != nil {
-		t.Fatalf("could not install chart-operator-resource-chart %v", err)
+		t.Fatalf("could not delete %q %v", cr, err)
 	}
 
-	var rc *helmclient.ReleaseContent
+	err = waitForReleaseStatus(gsHelmClient, testRelease, "DELETED")
+	if !helmclient.IsReleaseNotFound(err) {
+		t.Fatalf("%q not succesfully deleted %v", testRelease, err)
```
indyfree: Since we purge the release, we are actually not waiting for a "DELETED" Status but for a "release not found" error. Deletion takes a while, so we need to wait till the `waitForReleaseStatus`/backoff returns that error. Any ideas how to match that error faster are welcome.

indyfree:
E2e with merged delete fixes passes now. Opening for review.

integration/test/basic/basic_test.go
```diff
@@ -27,39 +32,55 @@ func TestChartInstalled(t *testing.T) {
...
+
+	err = waitForReleaseStatus(gsHelmClient, testRelease, "DEPLOYED")
+	if err != nil {
+		t.Fatalf("could not get release status of %q %v", testRelease, err)
+	}
+	log.Printf("%q succesfully deployed", testRelease)
+
+	// Test Deletion
+	log.Printf("deleting %q", cr)
+	err = helmClient.DeleteRelease(cr)
```
indyfree: Note: Here we are using the regular helmClient, since `chart-operator-resources` are installed with the normal tiller.

integration/test/basic/basic_test.go
MarcelMue: You could write a new functions which checks for the specific error in each backoff. 
This here should be fine though.

integration/test/basic/basic_test.go
```diff
@@ -27,39 +32,55 @@ func TestChartInstalled(t *testing.T) {
...
+	// Test Creation
+	log.Printf("creating %q", cr)
+	err = f.InstallResource(cr, templates.ChartOperatorResourceValues, ":stable")
+	if err != nil {
+		t.Fatalf("could not install %q %v", cr, err)
+	}
+
+	err = waitForReleaseStatus(gsHelmClient, testRelease, "DEPLOYED")
+	if err != nil {
+		t.Fatalf("could not get release status of %q %v", testRelease, err)
```
MarcelMue: This was here before but the text is slightly misleading IMO.
The problem here is that the status did not get set to `DEPLOYED`, right?

indyfree: Yes exactly. The release has to get deployed first before we can retrieve a status. The output looks during the tests looks like this:
```
018/05/08 15:27:12 getting release status 3.36370905s: could not retrieve release content: tb-release: release not found
2018/05/08 15:27:16 getting release status 10.578945892s: could not retrieve release content: tb-release: release not found
2018/05/08 15:27:26 getting release status 6.521996113s: waiting for "DEPLOYED", current "PENDING_INSTALL"
``` 
I think it's fine, but if you find a better wording to make it clearer I am open.

integration/test/basic/basic_test.go
indyfree: Yes thought about it too, but felt too much like duplicate code, just to fail one layer earlier. Think I am going with this.

indyfree:
Adressed reviews. PTAL :)

PR #66, opened 2018-05-08
Fix resource deletion

indyfree:

Fixes a issue when `chart-operator` doesn't recognize the chart to delete because `ReleaseStatus` is not carried in the desired State. So now we only compare relevent fields using an `Equals` method as discussed in Slack and other issues.

Also purges the release, so we don't have it in the current state anymore.

E2e test is here: #63 , but still failing because of other rbac/tiller issues that are beeing solved atm.


service/controller/v1/resource/chart/delete.go
```diff
@@ -3,33 +3,28 @@ package chart
...
+
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/operatorkit/controller"
 )
 
 func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange interface{}) error {
-	customObject, err := key.ToCustomObject(obj)
-	if err != nil {
-		return microerror.Mask(err)
-	}
```
indyfree: Boyscouting: We actually don't need the custom object here, since we can delete by ReleaseName of the `deleteChange`.

service/controller/v1/resource/chart/delete.go
```diff
@@ -3,33 +3,28 @@ package chart
...
 	chartState, err := toChartState(deleteChange)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
 	if chartState.ReleaseName != "" {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleting release %s", chartState.ReleaseName))
-		release := key.ReleaseName(customObject)
 
-		err := r.helmClient.DeleteRelease(release)
```
indyfree: See above: customObject not needed.

service/controller/v1/resource/chart/spec.go
```diff
@@ -21,3 +21,11 @@ type ChartState struct {
 	// e.g. 0.1.2
 	ReleaseVersion string
 }
+
+// Equals asseses the equality of ChartStates with regards to distinguishing fields
+func (a *ChartState) Equals(b ChartState) bool {
```
indyfree: Didn't find a better place to put the `Equals` method. Feedback welcome.

service/controller/v1/resource/chart/spec.go
```diff
@@ -21,3 +21,11 @@ type ChartState struct {
 	// e.g. 0.1.2
 	ReleaseVersion string
 }
+
+// Equals asseses the equality of ChartStates with regards to distinguishing fields
```
fgimenez: Nitpick, missing period at the end of godoc.

service/controller/v1/resource/chart/spec.go
```diff
@@ -21,3 +21,11 @@ type ChartState struct {
...
+func (a *ChartState) Equals(b ChartState) bool {
+	if a.ReleaseName != b.ReleaseName || a.ReleaseVersion != b.ReleaseVersion || a.ChartName != b.ChartName {
```
fgimenez: Maybe just `return a.ReleaseName == b.ReleaseName && a.ReleaseVersion == b.ReleaseVersion && a.ChartName == b.ChartName`?

service/controller/v1/resource/chart/delete.go
```diff
@@ -3,33 +3,28 @@ package chart
...
+		err := r.helmClient.DeleteRelease(chartState.ReleaseName, helm.DeletePurge(true))
```
fgimenez: Good to have purge in place.

service/controller/v1/resource/chart/spec.go
fgimenez: I think it's ok to have it here close to `ChartState` type definition :+1:

service/controller/v1/resource/chart/delete.go
```diff
@@ -60,7 +55,7 @@ func (r *Resource) newDeleteChange(ctx context.Context, obj, currentState, desir
 
 	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding out if the %s release has to be deleted", desiredChartState.ReleaseName))
 
-	if !reflect.DeepEqual(currentChartState, ChartState{}) && reflect.DeepEqual(currentChartState, desiredChartState) {
+	if !currentChartState.Equals(ChartState{}) && currentChartState.Equals(desiredChartState) {
```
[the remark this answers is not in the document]
rossf7: `IsEmpty` is nice for readability.

indyfree:
Added `IsEmpty` and went for seperate conditions in `IsEqual` for readibility. Waiting for e2e-deletion tests before I merge though.

indyfree:
Successfully e2e tested with #63. Merging after its own e2e run.

PR #68, opened 2018-05-10
vendor: updated helmclient

rossf7:
Updating helmclient as it has better error handling.


rossf7:
CI green. Vendor update. Merging.

PR #76, opened 2018-06-25
Add version endpoint and enable v2 resource set

rossf7:

Setting values in Charts will be implemented in v2.


server/endpoint/endpoint.go
```diff
@@ -2,6 +2,7 @@ package endpoint
 
 import (
 	"github.com/giantswarm/microendpoint/endpoint/healthz"
+	"github.com/giantswarm/microendpoint/endpoint/version"
```
rossf7: While enabling the v2 resource set I noticed that the version endpoint was not set up.

Even though chart-operator currently runs in guest clusters I think it should have the endpoint. Also later when we run it in the host cluster it will be needed.

service/controller/chart.go
```diff
@@ -136,12 +137,30 @@ func newChartResourceRouter(config ChartConfig) (*controller.ResourceRouter, err
 		}
 	}
 
+	var resourceSetV2 *controller.ResourceSet
```
rossf7: Here the new resource set is wired up.

rossf7:
The handlesFunc in the resource sets wasn't implemented. CI is green now. Going with this.

PR #78, opened 2018-06-26
Set values when installing or updating charts

rossf7:

### DIFF
--- a/service/controller/v2/resource/chart/create.go
+++ b/service/controller/v2/resource/chart/create.go
@@ -2,6 +2,7 @@ package chart
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


service/controller/v2/resource/chart/create.go
```diff
@@ -53,7 +59,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		//
 		err = r.helmClient.InstallFromTarball(tarballPath, ns,
 			helm.ReleaseName(chartState.ReleaseName),
-			helm.ValueOverrides([]byte("{}")),
+			helm.ValueOverrides(values),
```
fgimenez: <3

Issue #73, opened 2018-06-13
chart-operator locks if helm chart can not create all pods


r7vme:
I had a case, when cluster has only one worker and chart-operator stuck on installing nginx-ingress-controller chart, because two of three pods were pending.
```
{"caller":"github.com/giantswarm/chart-operator/service/controller/v1/resource/chart/create.go:24","event":"update","function":"ApplyCreateChange","level":"debug","message":"creating chart kubernetes-nginx-ingress-controller-chart","object":"/apis/core.giantswarm.io/v1alpha1/namespaces/giantswarm/chartconfigs/kubernetes-nginx-ingress-controller-chart","resource":"chartv1","time":"2018-06-13T13:53:32.589055+00:00"}
```

I assume we need to change some `helm install` flags to not wait until all pods will be runnning.

r7vme:
cc: @fgimenez @rossf7

rossf7:
Damnit, yes we need to be able to pass values to Helm charts. This is because in k8scloudconfig we have templating for the number of workers.

See https://github.com/giantswarm/giantswarm/issues/3389

@puja108 I think we need to tackle this next on Managed Services.

r7vme:
Is it possible to install chart asynchronously?

One of the customers expects quick provisioning of the cluster services and additional saved 10 minutes would be great boost.

puja108:
yes, apart from the values support we are building we should look into making chart installs non-blocking, which also sounds like the more k8s way, the job of chart-operator is just sending it over to the K8s API through Tiller, then it's the job of K8s to make that desired state also the current state. Reconcilliation of chart-operator should be based on what k8s knows as desired state, and reconciliation of current state is k8s master territory.

rossf7:
Yes we should proceed to the next chart. I think part of the problem is we're waiting when installing the chart.

https://github.com/giantswarm/chart-operator/blob/2076511f3f54d3baabbfa6499617b5a483690765/service/controller/v1/resource/chart/create.go#L57

@fgimenez @indyfree Do you know why we did it this way?

fgimenez:
[...]

PR #79, opened 2018-06-27
remove InstallWait from create chart code

fgimenez:
Towards https://github.com/giantswarm/chart-operator/issues/73

No need to wait for chart installed, the operator can send the installation request to tiller and continue.

### DIFF
--- a/service/controller/v2/resource/chart/create.go
+++ b/service/controller/v2/resource/chart/create.go
@@ -59,8 +59,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		//
 		err = r.helmClient.InstallFromTarball(tarballPath, ns,
 			helm.ReleaseName(chartState.ReleaseName),
-			helm.ValueOverrides(values),
-			helm.InstallWait(true))
+			helm.ValueOverrides(values))
 		if err != nil {
 			return microerror.Mask(err)
 		}


service/controller/v2/resource/chart/create.go
```diff
@@ -59,8 +59,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		//
 		err = r.helmClient.InstallFromTarball(tarballPath, ns,
 			helm.ReleaseName(chartState.ReleaseName),
-			helm.ValueOverrides(values),
-			helm.InstallWait(true))
+			helm.ValueOverrides(values))
```
MarcelMue: Maybe an unlikely scenario but what will happen if the install takes longer than the reconciliation period? Will we tell helm to install again even though the last install is still going on?

fgimenez: Nope, we return a non-empty create change only if the name of the chart in current state is empty or the name of the chart in the desired state doesn't match the name in the current state, see https://github.com/giantswarm/chart-operator/blob/master/service/controller/v2/resource/chart/create.go#L89.

After sending a request to install tiller will create a release, initially in `PENDING_INSTALL` state. So, in the next reconciliation loop, when we query the current state of the chart here https://github.com/giantswarm/chart-operator/blob/master/service/controller/v2/resource/chart/current.go#L24 the release will be found, either yet in `PENDING_INSTALL` or already in `DEPLLOYED` or `FAILED` state. In any case, the name won't be empty, will match the name of the desired state and the create change will be empty.

rossf7: Nice, thanks for the explanation.

fgimenez:
CI green, merging.

PR #80, opened 2018-07-02
Only compare release version and chart values when updating

rossf7:
Fixes a bug with updating chart values.

### DIFF
--- a/service/controller/v2/resource/chart/resource.go
+++ b/service/controller/v2/resource/chart/resource.go
@@ -72,13 +72,15 @@ func (r *Resource) Name() string {
 }
 
 func isChartModified(a, b ChartState) bool {
-	// ChartValues have changed so we need to update.
-	if !reflect.DeepEqual(a.ChartValues, b.ChartValues) {
+	// ReleaseVersion has changed for the channel so we need to update the Helm
+	// Release.
+	if a.ReleaseVersion != b.ReleaseVersion {
 		return true
 	}
 
-	// ReleaseVersion has changed for the channel so we need to update.
-	if a.ReleaseVersion != b.ReleaseVersion {
+	// ChartValues have changed so we need to update the values for the current
+	// Helm Release.
+	if !reflect.DeepEqual(a.ChartValues, b.ChartValues) {
 		return true
 	}
 
--- a/service/controller/v2/resource/chart/resource.go
+++ b/service/controller/v2/resource/chart/resource.go
@@ -1,6 +1,8 @@
 package chart
 
 import (
+	"reflect"
+
 	"github.com/giantswarm/apprclient"
 	"github.com/giantswarm/helmclient"
 	"github.com/giantswarm/microerror"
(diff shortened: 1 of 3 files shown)


service/controller/v2/resource/chart/update.go
```diff
@@ -95,7 +94,7 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
 
 	r.logger.LogCtx(ctx, "level", "debug", "message", "finding out if the chart has to be updated")
 
-	if currentChartState.ReleaseVersion != "" && !reflect.DeepEqual(desiredChartState, currentChartState) {
```
rossf7: Bug meant this was always returning true. The release status is set for the current state but not the desired state.

service/controller/v2/resource/chart/resource.go
```diff
@@ -69,6 +71,21 @@ func (r *Resource) Name() string {
...
+func isChartModified(a, b ChartState) bool {
+	// ChartValues have changed so we need to update.
+	if !reflect.DeepEqual(a.ChartValues, b.ChartValues) {
+		return true
+	}
+
+	// ReleaseVersion has changed for the channel so we need to update.
+	if a.ReleaseVersion != b.ReleaseVersion {
+		return true
+	}
```
rossf7: In v1 we just reconcile on the ReleaseVersion. So this is now correct.

service/controller/v2/resource/chart/resource.go
```diff
@@ -69,6 +71,21 @@ func (r *Resource) Name() string {
 	return Name
 }
 
+func isChartModified(a, b ChartState) bool {
+	// ChartValues have changed so we need to update.
```
fgimenez: Nitpick, we need to update if the next `if` condition evaluates to true, is that correct?

service/controller/v2/resource/chart/update.go
fgimenez: Very good catch!

service/controller/v2/resource/chart/resource.go
```diff
@@ -69,6 +71,21 @@ func (r *Resource) Name() string {
...
+	if !reflect.DeepEqual(a.ChartValues, b.ChartValues) {
+		return true
+	}
+
+	// ReleaseVersion has changed for the channel so we need to update.
```
fgimenez: Same as above, the comment assumes that the check is going to be true.

service/controller/v2/resource/chart/update_test.go
```diff
@@ -34,7 +34,21 @@ func Test_Resource_Chart_newUpdateChange(t *testing.T) {
 			expectedUpdateState: nil,
 		},
 		{
-			description: "case 1: nonempty current state, different release version in desired state, expected desired state",
+			description: "case 1: nonempty current state, equal desired state, empty update change",
```
fgimenez: Nice to modify/extend existing tests without explicitly testing `isChartModified`, only its external effect :+1:

service/controller/v2/resource/chart/resource.go
rossf7: Good point. I'm going to make the ReleaseVersion check first. As if it has changed we need to update the Helm release with the new chart version. 

If the values have changed we just need to update them for the current Helm release. I'll update the comments to make them clearer.

rossf7: Done.

service/controller/v2/resource/chart/resource.go
rossf7: Done.

PR #83, opened 2018-07-05
Fixed registry for chart-operator-chart

fgimenez:
Uses quay.io as the registry for chart-operator, even though it is installed as a chart by cluster-operator, the build process is still dependent on quay.io.


kopiczko:
For my interest: so quay.io isn't blocked in china? Why do we need other registry at all in that case?

corest:
It is slow. Such slow, that services fail while waiting for images to be pulled

fgimenez:
Exactly, sometimes it goes better, in the worst cases for the hyperkube and etcd images retagged on quay the default timeout of 90secs for starting a systemd unit was not enough to download them, and the nodes kept retrying and retrying without success.

PR #97, opened 2018-07-18
Remove hold step

MarcelMue:
No reason to keep this hold step here. We should always run these and have enough circle builds in parallel now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,17 +91,9 @@ workflows:
     jobs:
       - build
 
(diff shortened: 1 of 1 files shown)


MarcelMue:
This works here, because we don't need to spin up a guest cluster to run e2e tests for `chart-operator`. I don't see these taking very long in the future.

MarcelMue:
Going with this we can still enable it back when we feel like it's too much.

PR #100, opened 2018-07-24
Add prometheus collector

fgimenez:
Exposes metrics about ChartConfigs, including status.

### DIFF
--- a/service/collector/chart_config.go
+++ b/service/collector/chart_config.go
@@ -1,0 +1,84 @@
+package collector
+
+import (
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	"github.com/prometheus/client_golang/prometheus"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	gaugeValue          float64 = 1
+	chartNameLabel              = "chart_name"
+	channelNameLabel            = "channel_name"
+	releaseNameLabel            = "release_name"
+	releaseStatusLabel          = "release_status"
+	releaseVersionLabel         = "release_version"
+)
+
+type chartState struct {
+	chartName      string
+	channelName    string
+	releaseName    string
+	releaseStatus  string
+	releaseVersion string
+}
+
+var (
+	chartConfigDesc *prometheus.Desc = prometheus.NewDesc(
+		prometheus.BuildFQName(Namespace, "", "chartconfig_status"),
+		"Managed charts status.",
+		[]string{
+			chartNameLabel,
+			channelNameLabel,
+			releaseNameLabel,
+			releaseStatusLabel,
+		},
+		nil,
+	)
+)
+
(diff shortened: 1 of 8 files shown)


fgimenez:
This is how the metric looks like:
```
# HELP chart_operator_chartconfig_status Managed charts status.
# TYPE chart_operator_chartconfig_status gauge
chart_operator_chartconfig_status{channel_name="5-5-beta",chart_name="tb-chart",release_name="tb-release",release_status="DEPLOYED"} 1
```

fgimenez:
I've also taken the opportunity to set `versionBundleVersion` value to chart-operator-resource chart in the tests to `0.3.0`, so that we test the current WIP version on CI instead of `v1`.

service/collector/chart_config.go
```diff
@@ -0,0 +1,82 @@
+package collector
+
+import (
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	"github.com/prometheus/client_golang/prometheus"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
rossf7: Nitpick but I think this should be `metav1` to match our usual convention.

service/collector/chart_config.go
```diff
@@ -0,0 +1,82 @@
...
+			channelNameLabel,
+			releaseNameLabel,
+			releaseStatusLabel,
+		},
+		nil,
+	)
+)
+
+func (c *Collector) collectChartConfigStatus(ch chan<- prometheus.Metric) {
+	c.logger.Log("level", "debug", "message", "collecting metrics for vpcs")
```
rossf7: `vpcs` looks wrong here.

service/collector/chart_config.go
```diff
@@ -0,0 +1,82 @@
...
+		)
+	}
+	c.logger.Log("level", "debug", "message", "finished collecting metrics for ChartConfigs")
+}
+
+func (c *Collector) getChartConfigs() ([]*chartState, error) {
+	r, err := c.g8sClient.CoreV1alpha1().
+		ChartConfigs("giantswarm").
+		List(v1.ListOptions{})
+
```
rossf7: I think the line break could go here.

service/collector/collector.go
```diff
@@ -0,0 +1,67 @@
...
+func (c *Collector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- chartConfigDesc
+}
+
+func (c *Collector) Collect(ch chan<- prometheus.Metric) {
+	c.logger.Log("level", "debug", "message", "collecting metrics")
+
+	collectFuncs := []func(chan<- prometheus.Metric){
+		c.collectChartConfigStatus,
+	}
```
rossf7: Good to use the updated approach here with the WaitGroup, As we may need to add more collector funcs later.

PR #104, opened 2018-07-30
v3/resource/chart: make create & update checks consistent

kopiczko:


service/controller/v3/resource/chart/create.go
```diff
@@ -85,7 +85,7 @@ func (r *Resource) newCreateChange(ctx context.Context, obj, currentState, desir
 
 	createState := &ChartState{}
 
-	if currentChartState.ChartName == "" || desiredChartState.ChartName != currentChartState.ChartName {
```
indyfree: Now we don't check if `desiredChartState.ChartName != ""` , but guess thats not needed or done in another place?

MarcelMue: Why do we no longer check the second condition though?  Should we still check if the ChartNames are different?

kopiczko: `ChartName:      key.ChartName(customObject),` This is how the `ChartName` is set in both current & desired state. Checking condition that `ChartName` changes confuses me. Correct me if I'm wrong but this resource simply doesn't support that case.

rossf7: Yes this resource just installs chart-operator via its chart. Since its only dealing with a single chart I think this is fine and a good simplification.

MarcelMue: Oh yeah, looking at it you seem right.

rossf7: ~~Since its only dealing with a single chart~~ I think this is fine and a good simplification.

I got this mixed up with cluster-operator. In chart-operator it is dealing with multiple charts but I still think this change is fine.

PR #107, opened 2018-07-30
Use templating from e2etemplates

MarcelMue:


integration/test/basic/basic_test.go
```diff
@@ -34,6 +34,15 @@ func TestChartLifecycle(t *testing.T) {
 		},
 	}
 
+	chartConfigValues := chartconfig.ChartConfigValues{
+		Channel:   "5-5-beta",
+		Name:      "tb-chart",
+		Namespace: "giantswarm",
+		Release:   "tb-release",
+		//TODO: fix this static VersionBundleVersion
```
kopiczko: Should we have a TODO issue for that?

integration/chartconfig/chartconfig.go
```diff
@@ -0,0 +1,43 @@
...
+	ResourceVersion string
+}
+
+type Secret struct {
+	Name            string
+	Namespace       string
+	ResourceVersion string
+}
+
+func (ccv ChartConfigValues) ExecuteChartValuesTemplate() (string, error) {
```
kopiczko: Again me and data structs. I really don't like having methods like that on pure data structs. I think this one here should be a standalone func. Otherwise we end up with adding things like logger to the data structs.

integration/chartconfig/chartconfig.go
```diff
@@ -0,0 +1,43 @@
...
+package chartconfig
+
+import (
+	"bytes"
+	"html/template"
+
+	"github.com/giantswarm/e2etemplates/pkg/e2etemplates"
+)
+
+type ChartConfigValues struct {
```
kopiczko: It would be cool if `e2etemplates` was providing that struct. Otherwise we'll have to define it in every project that uses the template.

integration/test/chartvalues/chartvalues_test.go
```diff
@@ -22,14 +22,27 @@ func TestChartValues(t *testing.T) {
...
 	}
 
 	// Test Creation
 	l.Log("level", "debug", "message", fmt.Sprintf("creating %s", cr))
-	err = f.InstallResource(cr, templates.ChartOperatorResourceValues, ":stable")
+	chartValues, err := chartConfigValues.ExecuteChartValuesTemplate()
+	if err != nil {
+		t.Fatalf("could not template chart values %q %v", chartValues, err)
+	}
+	err = r.InstallResource(cr, chartValues, "stable")
```
indyfree: <3

integration/test/chartvalues/main_test.go
```diff
@@ -16,6 +17,7 @@ var (
 	f          *framework.Host
 	helmClient *helmclient.Client
 	l          micrologger.Logger
+	r          *resource.Resource
```
indyfree: Why did `err = r.UpdateResource(cr, templates.UpdatedChartOperatorResourceValues, "stable")` [1] work when this wasn't added before?

[1] https://github.com/giantswarm/chart-operator/pull/107/files#diff-5c729e87953427aa6b0c419816241bc1L51

MarcelMue: Because you are looking at different tests :P 
Resource got added in the `chartvalues` test and already existed in the `basic` test.

integration/chartconfig/chartconfig.go
MarcelMue: I'll provide a PR to `e2etemplates` and we can discuss there.

integration/test/chartvalues/main_test.go
indyfree: :sweat_smile: haha thanks for the pointer.

Gopkg.toml
```diff
@@ -68,6 +68,10 @@ required = [
   branch = "master"
   name = "github.com/giantswarm/k8sportforward"
 
+[[constraint]]
+  branch = "master"
+  name = "github.com/giantswarm/e2etemplates"
+
```
rossf7: Nit but I'd like the ordering to be alphabetical.

integration/test/basic/main_test.go
```diff
@@ -57,14 +57,16 @@ func init() {
...
+	{
+		c := resource.ResourceConfig{
+			Logger:     l,
+			HelmClient: helmClient,
+			Namespace:  "giantswarm",
+		}
+		r, err = resource.New(c)
+		if err != nil {
+			panic(err.Error())
+		}
```
rossf7: <3

integration/test/chartvalues/chartvalues_test.go
```diff
@@ -7,7 +7,8 @@ import (
 	"testing"
 
 	"github.com/giantswarm/chart-operator/integration/chart"
-	"github.com/giantswarm/chart-operator/integration/templates"
+	"github.com/giantswarm/chart-operator/integration/chartconfig"
+	"github.com/giantswarm/e2etemplates/pkg/e2etemplates"
```
rossf7: Should be in a separate block.

Gopkg.toml
MarcelMue: Should be alphabetical now.

integration/test/basic/basic_test.go
MarcelMue: I will do this imminently, so a TODO issue is overhead IMO.

integration/chartconfig/chartconfig.go
MarcelMue: I moved the data structure to the e2etemplates repo. PTAL.

integration/test/chartvalues/chartvalues_test.go
MarcelMue: Done.

PR #110, opened 2018-08-01
Improve test setup code

rossf7:
Follow up for https://github.com/giantswarm/chart-operator/pull/103

Use init function and use existing logger for appr client.


integration/test/chartvalues/main_test.go
```diff
@@ -57,5 +55,10 @@ func TestMain(m *testing.M) {
 		}
 	}
```
indyfree: Nitpick: Think theres an excessive whiteline.

PR #111, opened 2018-08-01
Use env package for env vars

rossf7:


MarcelMue:
Here we go with the clash, was just working on this haha.

integration/env/common.go
```diff
@@ -0,0 +1,32 @@
+package env
```
rossf7: Just getting this in line with the current approach.

integration/env/common.go
```diff
@@ -0,0 +1,32 @@
...
+	"os"
+)
+
+const (
+	// EnvVarCircleCI is the process environment variable representing the
+	// CIRCLECI env var.
+	EnvVarCircleCI = "CIRCLECI"
+	// EnvVarKeepResources is the process environment variable representing the
+	// KEEP_RESOURCES env var.
+	EnvVarKeepResources = "KEEP_RESOURCES"
```
MarcelMue: There are some others like `CIRCLE_SHA1`, that I want to get done - but this works as a start.

rossf7:
@MarcelMue Oops! We're treading on each others toes after all! ;)

I can close in favor of your PR if you've already done this?

integration/env/common.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+var (
+	circleCI      string
+	keepResources string
+)
+
+func init() {
+	circleCI = os.Getenv(EnvVarCircleCI)
```
MarcelMue: In `aws-operator` we panic if env vars aren't set. I am not a massive fan of this, but it works since this is an `init` func.
Unsure if we want to keep doing that.

rossf7: I don't think we can panic because the env vars just might not be set. e.g. when running locally.

We now panic if the Circle SHA is an empty string as that is an error condition we can trap.

integration/env/common.go
rossf7: I didn't add `CIRCLE_SHA1` because it isn't used yet. But I've added because we will need it. Also I think it makes sense to standardise this code.

rossf7:
@MarcelMue I've added the Circle SHA support. I'll merge this provided there are no objections.

integration/env/common.go
MarcelMue: Gotcha, cool & thanks :+1:

MarcelMue:
Mhm, e2e is flapping? I had a flap earlier but thought it was a one time fluke. Did we change anything major lately?

rossf7:
Not that I know of. Circle is still reporting GitHub issues even though GH have said everything is back. Re-running the test. Let's keep an eye on it.

MarcelMue:
Still failing? What can we do here to fix this?

rossf7:
Yes still failing but I've been working on other things. I'll re-run and try and debug it now.

It's failing intermittently. So not sure if its this change or another recent one.

integration/test/basic/basic_test.go
```diff
@@ -59,6 +59,12 @@ func TestChartLifecycle(t *testing.T) {
 		t.Fatalf("could not install %q %v", cr, err)
 	}
 
+	err = r.WaitForStatus(cr, "DEPLOYED")
+	if err != nil {
+		t.Fatalf("could not get release status of %q %v", cr, err)
+	}
+	l.Log("level", "debug", "message", fmt.Sprintf("%s succesfully deployed", cr))
```
rossf7: @MarcelMue FYI. I added a wait for the e2e chart being deployed, before the wait for the test chart being deployed. 

I think this makes the test more reliable. Re-running the test. If it succeeds then I'll merge.

indyfree: <3

PR #112, opened 2018-08-02
Use indexblob to determine versionbundle version

MarcelMue:
This uses the versionbundle version from the indexblob in our e2e tests.


integration/env/common.go
```diff
@@ -31,6 +44,34 @@ func init() {
...
+	}
+
+	var err error
+	params := &framework.VBVParams{
+		Component: "chart-operator",
+		Provider:  "aws",
+		Token:     token,
+		VType:     TestedVersion(),
+	}
+	versionBundleVersion, err = framework.GetVersionBundleVersion(params)
```
MarcelMue: This is aligned to `aws-operator` right now. I want to move this stuff out of env in the future, possibly into setup?

integration/test/basic/basic_test.go
```diff
@@ -33,12 +34,11 @@ func TestChartLifecycle(t *testing.T) {
...
-		Name:      "tb-chart",
-		Namespace: "giantswarm",
-		Release:   "tb-release",
-		//TODO: fix this static VersionBundleVersion
-		VersionBundleVersion: "0.2.0",
+		Channel:              "5-5-beta",
+		Name:                 "tb-chart",
+		Namespace:            "giantswarm",
+		Release:              "tb-release",
+		VersionBundleVersion: env.VersionBundleVersion(),
```
MarcelMue: Now using the version out of the `versionblob`.

.circleci/config.yml
```diff
@@ -41,24 +41,52 @@ jobs:
         - ./chart-operator
         - ./architect
 
-  e2eTestBasicPR:
+  e2eTestCurBasicPR:
```
MarcelMue: Diff looks a bit strange, but I added the jobs aligned with `aws-operator` here.

integration/test/basic/basic_test.go
rossf7: <3

PR #114, opened 2018-08-09
Add namespace to ChartConfig metrics

fgimenez:
This is required for preventing the guest cluster metrics to be dropped.


rossf7:
Changes LGTM but I know why e2e is borked :(

We added chart-operator to the indexBlob but I had to remove because chart-operator isn't deployed in host clusters yet. I removed from installations but I didn't sort here.

On it. I already created a vendor PR earlier but I didn't get a chance to investigate the e2e failures.

fgimenez:
Ok, the changes for getting versions from the indexblob were reverted on #115, I also took the opportunity to fix e2e tests there. CI green here after merging master, merging.

PR #117, opened 2018-08-15
Push chart to quay with sha version

MarcelMue:
I talked about this with @rossf7 and @JosephSalisbury to try and get `chart-operator` deployed through `draughtsman`.

The idea here is to replace the version which is used for `chart-operator` on guest clusters  with the usual `1.0.0-SHA` version which is required for draughtsman.

towards https://github.com/giantswarm/giantswarm/issues/3907

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,10 +74,9 @@ jobs:
     - run:
         name: publish to beta
         command: ./architect publish
-
-    - run: sed -i 's/version:.*/version: 1.0.0-'"${CIRCLE_SHA1}"'/' helm/chart-operator-chart/Chart.yaml
-
-    - run: ./architect deploy
+    - run: |
+        sed -i 's/version:.*/version: 1.0.0-'"${CIRCLE_SHA1}"'/' helm/chart-operator-chart/Chart.yaml
+        ./architect deploy
 
   publish_to_stable:
     machine: true
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,8 +74,9 @@ jobs:
     - run:
         name: publish to beta
         command: ./architect publish
-        
+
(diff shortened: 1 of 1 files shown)


kopiczko:
Are we going to deploy chart-operator with draughtsman long term? Shouldn't be chart-operator installed instead of draughtsman?

fgimenez:
hmm not sure if this is needed, why not just:
- point kubectl to _control plane_ ;)
- `helm registry install quay.io/giantswarm/chart-operator-chart@<version>`

If you want to install the latest version from stable you can use also the stable clannel, as we do on e2e tests:

`helm registry install quay.io/giantswarm/chart-operator-chart:stable`

AIUI this is the same approach we will take with `gsctl deploy` (without the kubectl setup in that case of course, just apprclient and helmclient would be required), but for control plane instead of tenant clusters, we might want to extend the changes to cover control plane too.

MarcelMue:
This is a short term solution:
Draughtsman is used for now and we plan to later on have `chart-operator`  manage itself.  As far as I understood, `chart-operator` does not manage itself so far and has its version managed by `cluster-operator` so far.

@fgimenez I don't think that is practical, the goal here is to have `chart-operator` on host cluster and be able to have it in the `indexblob` again. I do not want to manually apply `chart-operator` to all host clusters all the time. Or did I misunderstand what you want to do?

fgimenez:
Yes, I was talking about a more long-term solution, at some point bootstrapping chart-operator on control plane needs to be addressed. IMO the approach in this PR is something that will need to be changed sooner or later, but no blockers from my side.

MarcelMue:
I tested this again on godsmack one more time and it seems to be fine there as well. 

We will iterate on this in the (near) future but I want to get the ball rolling with `chart-operator` on the control plane.

PR #120, opened 2018-08-20
Add tiller reachable to collector

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/3925


service/collector/tiller_unreachable.go
```diff
@@ -0,0 +1,42 @@
...
+		c.logger.Log("level", "error", "message", "could not ping Tiller", "stack", fmt.Sprintf("%#v", err))
+
+		ch <- prometheus.MustNewConstMetric(
+			tillerUnreachableDesc,
+			prometheus.GaugeValue,
+			gaugeValue,
+			err.Error(),
+			defaultNamespace,
+		)
+	}
```
kopiczko: Shouldn't be `0` emitted in `else` clause?

service/collector/tiller_unreachable.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	errorLabel = "error"
+)
+
+var (
+	tillerUnreachableDesc *prometheus.Desc = prometheus.NewDesc(
+		prometheus.BuildFQName(Namespace, "", "tiller_unreachable"),
```
JosephSalisbury: i would invert this, actually. so something like

`chart_operator_tiller_up`, with a 1 if it is reachable, and a 0 if it's down

fgimenez: Done, instead of `tiller_up` I've named it `tiller_reachable`, this metric tells us about the state of tiller from chart-operator's point of view. I've also taken the opportunity to remove the error label, PTAL.

fgimenez:
> CI is red, guess still flapping?

yep, looks like it, all was green here https://circleci.com/workflow-run/08ccc76f-fcc1-4e35-86c4-630ff659ac9f

fgimenez:
Successfully tested on gauss, going with this.

PR #133, opened 2018-09-12
Set priority class for chart-operator pod

rossf7:


helm/chart-operator-chart/templates/deployment.yaml
```diff
@@ -27,6 +27,7 @@ spec:
           - key: config.yaml
             path: config.yaml
       serviceAccountName: {{ .Values.name }}
+      priorityClassName: system-cluster-critical
```
rossf7: I used the same priority as Ingress Controller. As we want the operator running so it can apply changes.

PR #134, opened 2018-09-14
Update microkit dep & enable debug server

tuommaki:
Added debug server exposes net/http/pprof interface in
http://127.0.0.1:6060/debug.

This helps with profiling running service e.g. when it's OOMing.

Towards https://github.com/giantswarm/giantswarm/issues/4088


tuommaki:
> Can you check why E2E failed? Or atleast create an issue about it? Seems to be something with the tiller pod.

Can't really find more than error that `chart-operator` cannot connect to tiller which then leads to timeout. Where should I create an issue for this and what kind of (i.e. which labels etc.)?

MarcelMue:
Just a normal issue on the releng board with a link to the build and so on. I want to have a look later and see if there is something we can do.

PR #138, opened 2018-09-19
Align rbac rules with other operators

MarcelMue:
We gave too many rights to the status resource foo. 
This is now aligned to e.g. aws-operator https://github.com/giantswarm/aws-operator/blob/master/helm/aws-operator-chart/templates/01-rbac.yaml#L26-L41


MarcelMue:
CI green, merging.

PR #140, opened 2018-09-25
Remove priority class

rossf7:
I added the priority class but its not working correctly in the `giantswarm` namespace.

```
Warning  FailedCreate  7m (x18 over 18m)  replicaset-controller  Error creating: pods "chart-operator-7c6db97676-" is forbidden: pods with system-cluster-critical priorityClass is not permitted in giantswarm namespace
```

Removing to fix alerts. I'll reintroduce once tested in the `giantswarm` namespace.

https://gigantic.slack.com/archives/C03CPNRTS/p1537860674000100

### DIFF
--- a/helm/chart-operator-chart/Chart.yaml
+++ b/helm/chart-operator-chart/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for the chart-operator
 name: chart-operator-chart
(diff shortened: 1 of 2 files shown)


rossf7:
2 e2e jobs still running but going with this to clear the alerts. I'll check on the merge to master as we may still have e2e problems.

PR #151, opened 2018-11-14
set tiller namespace to kube-system

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/4537

Point chart-operator to tiller in `kube-system` namespace on control planes.

### DIFF
--- a/helm/chart-operator-chart/templates/configmap.yaml
+++ b/helm/chart-operator-chart/templates/configmap.yaml
@@ -18,7 +18,7 @@ data:
       cnr:
         address: '{{ .Values.cnr.address }}'
       helm:
-        tillerNamespace:  '{{ .Values.namespace }}'
+        tillerNamespace:  '{{ .Values.tillerNamespace }}'
       kubernetes:
         incluster: true
         watch:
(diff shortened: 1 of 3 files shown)


kopiczko:
Did it change recently? How could it work before? For how long was it broken?

MarcelMue:
We weren't using `chart-operator` on control planes actively yet. It was added for the release-operator story which is currently in some teams backlog.

I am not sure how this works with tenant clusters with your change. This will break tenant usage of tiller?

TheoBrigitte:
chart-operator is broken on control plane, but it seems that we do not use it there. We only use chart-operator for tenant cluster where everything is fine.

MarcelMue:
Tiller on tenants is running in the `giantswarm` namespace because customers have a separate tiller in `default` or something like that. I do not see how this would work in this PR.

TheoBrigitte:
This PR change default tiller namespace to `kube-system` for control plane, another PR (https://github.com/giantswarm/cluster-operator/pull/329) set tiller namespace to `giantswarm` for tenant clusters.

I went this way because there is no easy solution to pass different values to charts being deployed to control plane.

TheoBrigitte:
e2e tests are failing since default namespace is kube-system now, I merge and deploy cluster-operator first, this should fix it.

PR #157, opened 2018-11-23
reconcile chart status

TheoBrigitte:
Towards: 
- https://github.com/giantswarm/giantswarm/issues/4077
- https://github.com/giantswarm/giantswarm/issues/4602

Reconcile charts status, and recover from failed charts deployment.

@indyfree and me played around with helm, and issuing a new upgrade/update command allow to recover from failed chart.


service/controller/v5/resource/chart/resource.go
```diff
@@ -84,6 +86,11 @@ func isChartModified(a, b ChartState) bool {
 		return true
 	}
 
+	// ReleaseStatus do not match so we need to update the Helm Release.
+	if a.ReleaseStatus != b.ReleaseStatus {
```
indyfree: We should also check what are the implications when installating the helm release and status == `PENDING`.

indyfree:
Ah also we have to remember to increase the `version` in Chart.yaml

service/controller/v5/resource/chart/resource.go
TheoBrigitte: good point, we have to handle state transition. so let's move this logic outside this function.

JosephSalisbury:
@TheoBrigitte is the idea here to update the CR status if the helm release is failed?

TheoBrigitte:
No, the idea here is to reconcile charts which have failed to be deployed.

service/controller/v5/resource/chart/update.go
```diff
@@ -51,7 +51,7 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
 			return microerror.Mask(err)
 		}
 
-		err = r.helmClient.UpdateReleaseFromTarball(ctx, releaseName, tarballPath, helm.UpdateValueOverrides(values))
+		err = r.helmClient.UpdateReleaseFromTarball(ctx, releaseName, tarballPath, helm.UpdateValueOverrides(values), helm.UpgradeForce(true))
```
TheoBrigitte: Didn't found a better solution than force upgrade in order to recover failed charts.

This might potentially delete/recreate resource. I haven't tested this.

pipo02mix: IMO this is dangerous because [it means could delete the chart and installed it again](https://github.com/helm/helm/issues/4177#issuecomment-407463879). For some charts like nginx controller it would have an impact for customer workloads. Are we sure?

cc/ @puja108

rossf7: Late to the party but agree with @pipo02mix that `upgrade --force` is dangerous for cluster critical apps. @TheoBrigitte Let me know if I can help here.

For me the failed releases is the biggest support problem with chart-operator. We need a solution and @giantswarm/team-batman will be working on chart-operator a lot for the managed apps story.

TheoBrigitte: Yeah I agree force upgrade is quite dangerous here and there might be some cases of failed chart which do not even need this.

I feel like the logic for state transition is something that would be needed to implement the force upgrade (or another recovery solution) later on. WDYT if we just go with state transition here ?

rossf7: @TheoBrigitte Thanks, going with the state transition logic for now is a good idea. 

I'll take care of reconciling the chart status as part of the App Catalog story. But we're still discussing how to configure the `upgrade --force` support in the spec PR.

TheoBrigitte:
Closing in favor of https://github.com/giantswarm/chart-operator/pull/178

PR #161, opened 2018-12-14
remove healthz service from operator

tomahawk28:
During https://github.com/giantswarm/app-operator/pull/3, we found out that we need to remove `healthz` service in use.


tomahawk28:
oops, forgot it, thanks

PR #162, opened 2018-12-19
Call EnsureTillerInstalled on boot

rossf7:
Towards giantswarm/giantswarm#4938

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,8 +1,8 @@
 package main
 
 import (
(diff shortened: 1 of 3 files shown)


rossf7:
Closing. I was being dumb. chart-operator already calls `EnsureTillerInstalled` its just there are no chartconfig CRs in control plane clusters yet.

PR #171, opened 2019-01-02
Add chart controller

rossf7:

Adds 2nd chart controller that handles the new chart CRD.


service/controller/chart/v1/key/key.go
```diff
@@ -0,0 +1,33 @@
+package key
+
+import (
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/microerror"
+)
+
+const (
+	versionBundleAnnotation = "giantswarm.io/version-bundle"
```
rossf7: Since the chart CRs are customer facing we use this annotation instead of having it under Spec.

See https://github.com/giantswarm/fmt/pull/36.

service/controller/chart/v1/resource/release/error.go
```diff
@@ -0,0 +1,12 @@
+package release
```
rossf7: I've renamed the `chart` resource to `release`. I think its a better name since the resource creates Helm releases from chart CRs.

service/controller/chart/v1/version_bundle.go
```diff
@@ -0,0 +1,20 @@
...
+		Changelogs: []versionbundle.Changelog{
+			{
+				Component:   "chart-operator",
+				Description: "Add your changes here.",
+				Kind:        versionbundle.KindAdded,
+			},
+		},
+		Components: []versionbundle.Component{},
+		Name:       "chart-operator",
+		Version:    "1.0.0",
```
rossf7: The operator needs to expose both chart and chartconfig version bundles. I don't want to add the filtering we have in cluster-operator because its complex and I don't think its needed.

`chart` version bundles will be `1.y.z` and `chartconfig` version bundles will be `0.y.z`.

service/controller/chartconfig/v5/resource_set.go
```diff
@@ -19,12 +19,6 @@ import (
 	"github.com/giantswarm/chart-operator/service/controller/chartconfig/v5/resource/chartstatus"
 )
 
-const (
-	// ResourceRetries presents number of retries for failed Resource
-	// operation before giving up.
-	ResourceRetries uint64 = 3
-)
-
```
rossf7: Removing dead code.

service/controller/chartconfig/v5/resource_set.go
```diff
@@ -46,6 +40,9 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 	var err error
 
 	// Dependencies.
+	if config.Fs == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Fs must not be empty", config)
+	}
```
rossf7: Just making the error handling consistent for both controllers.

service/controller/chart/chart.go
```diff
@@ -0,0 +1,131 @@
...
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	var newInformer *informer.Informer
+	{
+		c := informer.Config{
+			Logger:  config.Logger,
+			Watcher: config.G8sClient.ApplicationV1alpha1().Charts(config.WatchNamespace),
```
rossf7: Uses the new CRD.

rossf7:
@tomahawk28 We name the operatorkit resources after the resources they manage. So app-operator will take the app CRs and create chart CRs for chart-operator to manage.

chart-operator doesn't create any native k8s resources directly. It creates Helm releases when it deploys the charts. So I think the new name is more accurate. Does that make sense?

PR #174, opened 2019-01-10
Update vendor and remove mocks

rossf7:
Removes all the mock_test.go files and updates the unit tests. This means we can vendor helmclient changes without needing to update all the mocks.


service/controller/chart/v1/resource/release/current_test.go
```diff
@@ -108,10 +108,14 @@ func Test_CurrentState(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			helmClient := &helmMock{
-				defaultReleaseContent: tc.releaseContent,
-				defaultReleaseHistory: tc.releaseHistory,
-				defaultError:          tc.returnedError,
+			var helmClient helmclient.Interface
```
rossf7: This is all just using apprclienttest and helmclienttest in the unit tests.

PR #185, opened 2019-01-25
Add EnsureTillerInstalled to status resource

rossf7:
I hit errors running chart-operator against a fresh cluster with no tiller installed. Adds a call to the status resource for `EnsureTillerInstalled` to fix this.

### DIFF
--- a/service/controller/chart/v1/resource/status/create.go
+++ b/service/controller/chart/v1/resource/status/create.go
@@ -21,6 +21,11 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	releaseName := key.ReleaseName(cr)
 	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("getting status for release %#q", releaseName))
 
+	err = r.helmClient.EnsureTillerInstalled(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
 	releaseContent, err := r.helmClient.GetReleaseContent(ctx, releaseName)
 	if helmclient.IsReleaseNotFound(err) {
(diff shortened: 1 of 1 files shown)


service/controller/chart/v1/resource/status/create.go
```diff
@@ -21,6 +21,11 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	releaseName := key.ReleaseName(cr)
 	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("getting status for release %#q", releaseName))
 
+	err = r.helmClient.EnsureTillerInstalled(ctx)
```
xh3b4sd: We should check how we can improve the implementation of `EnsureTillerInstalled`. It does A LOT of stuff and uses two backoffs to wait on different stages. 

* https://github.com/giantswarm/helmclient/blob/c93791a1e9e60a08f2ee48394d416a726c5e00f2/helmclient.go#L265
* https://github.com/giantswarm/helmclient/blob/c93791a1e9e60a08f2ee48394d416a726c5e00f2/helmclient.go#L336

This makes `chart-operator` resources block for too long. Instead we should try to initialize some action quickly and come back the next time and check how far we got. We should not block and wait. Just init, stop, come back, check again.

xh3b4sd: screenshot 2019-01-23 at 22 19 27

xh3b4sd: https://github.com/giantswarm/giantswarm/issues/5182

rossf7: Ack. Makes sense that this is what is slowing down chart-operator. cluster-operator also does this and in tenant clusters which likely makes it worse.

Thanks for creating the issue. Going with this for now. But we should fix as sadly we'll need to run Tiller for a while yet.

PR #188, opened 2019-01-28
Move chartconfig e2e tests

rossf7:
I need to add e2e test coverage for the new chart CRD. This moves the existing tests into a `chartconfig` directory. More refactoring will come in later PRs as chart CRs do not need to use CNR.


integration/setup/setup.go
```diff
@@ -32,7 +32,7 @@ func WrapTestMain(ctx context.Context, h *framework.Host, helmClient *helmclient
 		v = 1
 	}
 
-	err = helmClient.EnsureTillerInstalled(context.TODO())
+	err = helmClient.EnsureTillerInstalled(ctx)
```
rossf7: Using the context already passed to `WrapTestMain`.

PR #189, opened 2019-01-29
Rename e2e setup to support chartconfig and chart CRs

rossf7:
This is prep for adding e2e for the new chart CRD controller. Adds the tested custom resource to the Circle config and renames some packages to make their purpose clearer.


integration/test/chartconfig/chartvalues/chartvalues_test.go
```diff
@@ -9,8 +9,8 @@ import (
 
 	"github.com/giantswarm/e2etemplates/pkg/e2etemplates"
 
-	"github.com/giantswarm/chart-operator/integration/chart"
-	"github.com/giantswarm/chart-operator/integration/chartconfig"
+	"github.com/giantswarm/chart-operator/integration/cnr"
+	"github.com/giantswarm/chart-operator/integration/customresource"
```
xh3b4sd: Wasn't the specific naming better when there are multiple different custom resources?

rossf7: `chart` is renamed to `cnr`. So yes that works. OK I'll go with that.

rossf7: Done. @xh3b4sd PTAL.

integration/env/common.go
```diff
@@ -44,33 +54,41 @@ func init() {
...
+	testedCustomResource = os.Getenv(EnvVarTestedCustomResource)
+	if testedCustomResource == "" {
+		panic(fmt.Sprintf("env var '%s' must not be empty", EnvVarTestedCustomResource))
 	}
 
-	if VersionBundleVersion() == "" {
-		if strings.ToLower(TestedVersion()) == "wip" {
-			log.Println("WIP version bundle version not present, exiting.")
-			os.Exit(0)
+	if testedCustomResource == ChartConfigCustomResource {
```
xh3b4sd: So instead of doing some hidden conditional routing in the code again, why don't we execute the test we want to execute? We could maybe have tests for the resources or functionality we want to test and isolate it from the rest which is not of interest. We define a test dir already, so this should maybe tell what to execute and the code would be cleaner.

.circleci/config.yml
```diff
@@ -53,48 +53,56 @@ jobs:
   e2eTestChartConfigCurBasicPR:
     environment:
       TEST_DIR: "integration/test/chartconfig/basic"
+      TESTED_CUSTOM_RESOURCE: "chartconfig"
```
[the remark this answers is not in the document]
kopiczko: Isn't `chartconfig` supposed to go away? I would just copy those tests. This is hard to maintain but are not supposed to support it for a long time. When we do the fix we have to apply it in two places anyway.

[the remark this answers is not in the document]
rossf7: chartconfig does go away but we use it in tenant clusters. I think we need coverage for both. We can remove the chartconfig tests when we remove the chartconfig controller.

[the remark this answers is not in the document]
kopiczko: Yes that's why I suggest to copy the chart tests and change chartconfig part and not really care about keeping them DRY as this is temporary state.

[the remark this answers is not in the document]
rossf7: Yeah OK I see what you mean. CNR specific logic is extracted from env. I'll do the same for setup and teardown.

integration/chartconfig/chartconfig.go
```diff
@@ -4,18 +4,128 @@ package chartconfig
...
+	return nil
+}
+
+func ExecuteValuesTemplate(ccv e2etemplates.ApiextensionsChartConfigValues) (string, error) {
 	buf := &bytes.Buffer{}
 	chartValuesTemplate := template.Must(template.New("chartConfigChartValues").Parse(e2etemplates.ApiextensionsChartConfigE2EChartValues))
 	err := chartValuesTemplate.Execute(buf, ccv)
 	if err != nil {
-		return "", err
+		return "", microerror.Mask(err)
```
xh3b4sd: <3

integration/chartconfig/chartconfig.go
```diff
@@ -4,18 +4,128 @@ package chartconfig
...
+
+	return nil
+}
+
+func VersionBundleVersion(githubToken, testedVersion string) (string, error) {
+	if githubToken == "" {
+		return "", microerror.Maskf(invalidConfigError, "env var %#q must not be empty", env.EnvVarGithubBotToken)
+	}
+	if testedVersion == "" {
+		return "", microerror.Maskf(invalidConfigError, "env var %#q must not be empty", env.EnvVarTestedVersion)
```
xh3b4sd: Let's please use these error types within constructors. Here the catch all would be `failedExecutionError`.

rossf7: Good catch. Done.

PR #196, opened 2019-02-12
Revert "Set priority class for chart-operator pod (#195)"

rossf7:
Removes the priority class. chart-operator runs in control planes but is unused. The priority classes are missing so this broke and triggered alerts.

The issue I was trying to fix is a problem reported by @pipo02mix with tenant clusters. Due to cluster-autoscaler the chart-operator pod can be evicted as it misses the priority class.

But we don't need a fix until cluster-autoscaler ships. I'll reintroduce once we have a fix that works for tenants and control planes.

### DIFF
--- a/helm/chart-operator-chart/templates/deployment.yaml
+++ b/helm/chart-operator-chart/templates/deployment.yaml
@@ -29,7 +29,6 @@ spec:
           items:
           - key: config.yaml
             path: config.yaml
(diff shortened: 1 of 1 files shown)


rossf7:
FYI @giantswarm/team-spirit we need a fix for this before cluster-autoscaler ships.

Can we just create the priority classes for control planes? Seems like it would be useful there too.

PR #201, opened 2019-02-20
vendor: updated deps

rossf7:
Vendor the latest helmclient changes.


MarcelMue:
@rossf7 How did you vendor this? `dep fails for me when I try to run `dep ensure` on master

rossf7:
@MarcelMue What is the error? I don't remember having dep problems.

MarcelMue:
@rossf7 I just created a new PR with a vendor update https://github.com/giantswarm/chart-operator/pull/203

The problem was that the `gopkg.in/yaml.v2` version was not compatible with the one from microkit.
It should be fixed now.

rossf7:
Ack. Thanks for the explanation. I added the YAML support fairly recently but I didn't encounter this problem.

PR #204, opened 2019-02-26
Default tiller namespace to kube-system

rossf7:
Fix for a bug found during integration testing yesterday. In the control plane we should use the existing tiller in kube-system and not create another tiller in giantswarm.

We only use tiller in giantswarm for tenant clusters to avoid conflicting with customers.

See https://github.com/giantswarm/cluster-operator/pull/398/files for the related change.

### DIFF
--- a/helm/chart-operator-chart/Chart.yaml
+++ b/helm/chart-operator-chart/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for the chart-operator
 name: chart-operator-chart
-version: 0.3.2
+version: 0.4.0
--- a/helm/chart-operator-chart/templates/configmap.yaml
+++ b/helm/chart-operator-chart/templates/configmap.yaml
@@ -18,7 +18,7 @@ data:
       cnr:
         address: '{{ .Values.cnr.address }}'
       helm:
-        tillerNamespace:  '{{ .Values.namespace }}'
+        tillerNamespace:  '{{ .Values.tiller.namespace }}'
       kubernetes:
         incluster: true
         watch:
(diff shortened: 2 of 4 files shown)


helm/chart-operator-chart/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for the chart-operator
 name: chart-operator-chart
-version: 0.3.2
+version: 0.4.0
```
rossf7: Bumps the version so we use a new release channel. This will be enabled in 0.12.0 of cluster-operator.

helm/chart-operator-chart/templates/configmap.yaml
```diff
@@ -18,7 +18,7 @@ data:
       cnr:
         address: '{{ .Values.cnr.address }}'
       helm:
-        tillerNamespace:  '{{ .Values.namespace }}'
+        tillerNamespace:  '{{ .Values.tiller.namespace }}'
```
rossf7: The new chart value.

helm/chart-operator-chart/values.yaml
```diff
@@ -10,6 +10,9 @@ replicas: 1
 
 clusterDNSIP: 172.31.0.10
 
+cnr:
+  address: https://quay.io
+
```
rossf7: Just ordering stuff.

helm/chart-operator-chart/values.yaml
```diff
@@ -23,5 +26,5 @@ resources:
     cpu: 250m
     memory: 250Mi
 
-cnr:
-  address: https://quay.io
+tiller:
+  namespace: kube-system
```
rossf7: Defaults to kube-system and we override to giantswarm when managed by cluster-operator.

PR #214, opened 2019-03-22
Add tiller configuration alert to collector

rossf7:
Towards giantswarm/giantswarm#5561

Adds metric `chart_operator_tiller_max_history` with value of `TILLER_HISTORY_MAX` in Tiller deployment. When this setting is incorrect we got too many tiller configmaps created.


rossf7:
Tested on ginger.

service/collector/key/key.go
```diff
@@ -0,0 +1,13 @@
...
+func TillerDeploymentName() string {
+	return "tiller-deploy"
+}
+
+func TillerMaxHistoryEnvVarName() string {
+	return "TILLER_HISTORY_MAX"
+}
+
+func TillerMaxHistoryEnvVarValue() string {
+	return "10"
```
tomahawk28:
```suggestion
	const (
	  TillerDeploymentName = "tiller-deploy"
	)
```
Does it make sense to use `const` rather than function?

rossf7: The key package uses the same approach we use in e2e. I don't think using constants adds much value.

https://github.com/giantswarm/aws-operator/blob/master/integration/key/release_name.go

rossf7:
@tomahawk28 Good shout on the PR desc. I'll update it.

PR #217, opened 2019-04-04
collector: fix log level

kopiczko:


service/collector/tiller_max_history.go
```diff
@@ -44,8 +44,8 @@ func (c *Collector) collectTillerConfigured(ctx context.Context, ch chan<- prome
 	if len(charts) == 0 && len(chartConfigs) == 0 {
 		// Skip checking tiller when there are no custom resources,
 		// as tiller is only installed when there is at least one CR to reconcile.
-		c.logger.Log("level", "message", "message", "did not collect Tiller max history")
-		c.logger.Log("level", "message", "message", "no Chart or ChartConfig CRs in the cluster")
+		c.logger.Log("level", "debug", "message", "did not collect Tiller max history")
+		c.logger.Log("level", "debug", "message", "no Chart or ChartConfig CRs in the cluster")
```
tuommaki: Would it make any sense to add log levels as functions in our logger?

kopiczko: I asked that question a while ago. I think @xh3b4sd is against that.

PR #220, opened 2019-04-17
Bump chart-operator helm chart to 0.6.0

rossf7:
We had some problems with the wrong commits getting pushed to the appr release channel. This is mitigated by using the 0-6-stable channel only in the WIP release.

See https://github.com/giantswarm/cluster-operator/pull/438


rossf7:
This is just a config change. Going with ping.

PR #222, opened 2019-04-18
Introduce reason into chartstatus resource

tomahawk28:
- Toward https://github.com/giantswarm/giantswarm/issues/5751
- Introduce new `reason` field into `chartstatus` resource.


tomahawk28:
Tested on `ginger`
```
...
status:
  reason: Upgrade complete
  releaseStatus: DEPLOYED
```

puja108:
```status:
  reason: Upgrade complete
  releaseStatus: DEPLOYED```

This looks a bit strange as "Upgrade complete" is not much of a "reason". "reason" might make sense in failure cases, but in success cases this sounds odd. @rossf7 I know you talked with @tomahawk28 a bit about what to call that field, but now that I see it in action I think we might wanna rethink the naming here to not confuse people.

**Comment by tomahawk28** [comment]
We have two options. 
1. We can display `reason` only when release status is `DEPLOYED`.
2. We can change the name of `reason` as `releaseDescription`, or just simply `description`

**Comment by rossf7** [comment]
Yes I agree. I think there are 2 options.

A: Rename the field from `reason` to `description` so its purpose is clear. For context here is the godoc for the field.

`// Description is a human-friendly "log entry" about this Helm release.`

B: Don't set the field if the release status is `DEPLOYED`. As `Upgrade complete` is not really useful information.

My preference would be option B. So we only show relevant information for debugging failed helm releases.

@puja108 @tomahawk28 WDYT?

**Comment by rossf7** [comment]
@tomahawk28 Oh wow our updates crossed. I like your option 1 but its actually the reverse.
 
We can display reason only when release status is **not** DEPLOYED.

**Comment by puja108** [comment]
I my testing I experienced a failure that was not connected to failed or deployed status, I could only see it in logs. Description would most probably catch that but making description or reason depending on status itself would not I guess? At least I wouldn't want to have a whitelist of stati but having a blacklist like if it's deployed then we're good that's ok. Latter is me assuming that once it's deployed we have no additional information about its status to show the user.

**Comment by rossf7** [comment]
> I my testing I experienced a failure that was not connected to failed or deployed status, I could only see it in logs.

It's a good improvement to have the reason in the CR status. But we should also see if we can improve the logs. I think often we show the raw error we get back from Tiller which has limited information.

IMO in that case we could also log the reason. So the logs show the why as well as what is happening.

**Comment by tomahawk28** [comment]
I think `helmclient` is a good point to start logging everything, better than individual operator. @puja108 WDYT?

**Comment by rossf7** [comment]
> I think helmclient is a good point to start logging everything, better than individual operator. 

Uncertain, chart-operator is our integration point with Tiller. But we do also use helmclient a lot in e2e.

Anyway this here improves the situation a lot. So I'd say we go with this and do further changes in later PRs.

File: service/controller/chartconfig/v5/resource/chartstatus/create.go [hunk 1]
```diff
@@ -34,6 +34,15 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 
 		customObjectCopy := customObject.DeepCopy()
 		customObjectCopy.Status.ReleaseStatus = releaseContent.Status
+
+		if releaseContent.Status != releaseStatusDeployed {
```
kopiczko: I'm not into details but what if the status becomes `DEPLOYED` again?

rossf7: Good catch thanks. I think because of the deep copy we need to clear the status when the release is deployed. Otherwise incorrect information will be displayed.

@tomahawk28 WDYT?

tomahawk28: Thanks a lot, I just make a change of it.

tomahawk28:
@kopiczko PTAL :)

PR #224, opened 2019-04-22
 Add g8sClient to release resource for updating the chart CR status

rossf7:
Towards giantswarm/giantswarm#4949

See https://github.com/giantswarm/apiextensions/pull/230 which adds an MD5 checksum to the chart CR status.

This adds the wiring to add a g8sClient to the release resource. It's prep for later PRs.


rossf7:
Not needed since we will use an annotation for the values checksum.

rossf7:
> Not needed since we will use an annotation for the values checksum.

Sorry I need to resurrect this. Whether we use the CR status or a CR annotation we still need a g8s client. 🤦‍♂️

PR #227, opened 2019-04-25
Merge configmap and secret values

rossf7:
Towards giantswarm/giantswarm#4949

This PR implements the merging of configmap and secrets using `helmclient.MergeValues`.

The values we get back from the Tiller API are not identical to the ones we provide. So an annotation 
`chart-operator.giantswarm.io/values-md5-checksum` is used for the comparison in the reconciliation logic.


service/controller/chart/v1/resource/release/desired.go
```diff
@@ -40,41 +42,59 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
+		valuesYAML, err = yaml.Marshal(values)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		valuesMD5Checksum = fmt.Sprintf("%x", md5.Sum(valuesYAML))
+	} else {
+		// We need to pass empty values in ValueOverrides to make the install
+		// process use the default values and prevent errors on nested values.
+		valuesYAML = []byte("{}")
```
rossf7: We still need to pass empty values if no values have been specified. This is so Helm uses the `values.yaml` for the chart.

service/controller/chart/v1/resource/release/desired_test.go
```diff
@@ -343,93 +348,20 @@ func Test_DesiredState(t *testing.T) {
 				t.Fatalf("error == %#v, want matching", err)
 			}
 
-			ReleaseState, err := toReleaseState(result)
+			releaseState, err := toReleaseState(result)
```
rossf7: Naming was wrong so sneaking in this fix.

service/controller/chart/v1/resource/release/resource.go
```diff
@@ -89,6 +93,39 @@ func (r *Resource) Name() string {
 	return Name
 }
 
+func (r *Resource) updateAnnotations(ctx context.Context, cr v1alpha1.Chart, releaseState ReleaseState) error {
```
tomahawk28: This function do 3 different things. 
1. Get the current Chart. 
2. Compare md5sum
3. Update `Chart` with a new one. 
But just name it `updateAnnotation` is a bit unclear for me. 
Can we divide the function of inside, or simply rename it?

rossf7: The comparison is so we only make a k8s API call if the annotation has changed. The reconciliation logic uses `isReleaseModified`. Should I add more comments to make that clearer?

service/controller/chart/v1/resource/release/resource.go
```diff
@@ -89,6 +93,39 @@ func (r *Resource) Name() string {
...
+		annotations[key.ValuesMD5ChecksumAnnotationName] = releaseState.ValuesMD5Checksum
+
+		currentCR.Annotations = annotations
+		_, err = r.g8sClient.ApplicationV1alpha1().Charts(cr.Namespace).Update(currentCR)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("updated annotations for app CR %#q in namespace %#q", cr.Name, cr.Namespace))
```
tomahawk28: This log only related to when we actually update `annotation`

service/controller/chart/v1/resource/release/resource.go
tomahawk28: Maybe some `godoc` for the function is enough then.

service/controller/chart/v1/resource/release/resource.go
rossf7: Good point. I'll fix the logging and add the missing godoc. Thx!

rossf7:
@tomahawk28 PTAL

PR #228, opened 2019-04-26
v5 Changelog

tomahawk28:
- add `status` change in v5 changelog

### DIFF
--- a/service/controller/chartconfig/v5/version_bundle.go
+++ b/service/controller/chartconfig/v5/version_bundle.go
@@ -9,7 +9,7 @@ func VersionBundle() versionbundle.Bundle {
 		Changelogs: []versionbundle.Changelog{
 			{
 				Component:   "chart-operator",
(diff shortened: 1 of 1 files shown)


service/controller/chartconfig/v5/version_bundle.go
```diff
@@ -9,7 +9,7 @@ func VersionBundle() versionbundle.Bundle {
 		Changelogs: []versionbundle.Changelog{
 			{
 				Component:   "chart-operator",
-				Description: "Add your changes here.",
+				Description: "Add reason field to chartconfig CR status when Helm release is in an error state.",
 				Kind:        versionbundle.KindAdded,
```
tomahawk28: What should it be, KindAdded? KindChange?

rossf7: Added is good as its a new field.

PR #236, opened 2019-05-29
 Enable force upgrade configured by annotation in both controllers

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/5745 and https://github.com/giantswarm/giantswarm/issues/5612

Adds a new `chart-operator.giantswarm.io/force-helm-upgrade` annotation to both chart and chartconfig CRs that triggers a "helm upgrade --force".

### DIFF
--- a/service/controller/chart/v1/key/key.go
+++ b/service/controller/chart/v1/key/key.go
@@ -1,11 +1,17 @@
 package key
 
 import (
+	"strconv"
+
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/microerror"
 )
 
 const (
+	// ForceHelmUpgradeAnnotationName is the name of the annotation that
+	// controls whether force is used when upgrading the Helm release.
+	ForceHelmUpgradeAnnotationName = "chart-operator.giantswarm.io/force-helm-upgrade"
+
 	// ValuesMD5ChecksumAnnotationName is the name of the annotation storing
 	// an MD5 checksum of the Helm release values.
 	ValuesMD5ChecksumAnnotationName = "chart-operator.giantswarm.io/values-md5-checksum"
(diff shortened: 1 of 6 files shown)


service/controller/chart/v1/resource/release/update.go
```diff
@@ -37,15 +37,16 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
 			}
 		}()
 
-		// TODO: Remove once upgrade --force support for re-installing releases
-		// is supported.
-		//
-		//	See https://github.com/giantswarm/giantswarm/issues/5612
-		//
```
rossf7: 🎉

service/controller/chart/v1/resource/release/update.go
```diff
@@ -37,15 +37,16 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
 
 		// We need to pass the ValueOverrides option to make the update process
 		// use the default values and prevent errors on nested values.
-		err = r.helmClient.UpdateReleaseFromTarball(ctx, releaseState.Name, tarballPath, helm.UpdateValueOverrides(releaseState.ValuesYAML))
+		err = r.helmClient.UpdateReleaseFromTarball(ctx, releaseState.Name, tarballPath,
+			helm.UpdateValueOverrides(releaseState.ValuesYAML),
+			helm.UpgradeForce(upgradeForce))
```
rossf7: I still need to test but AIUI we don't need to check the helm release status here. Resources will only be deleted first if the release is unhealthy.

> --force  force resource update through delete/recreate if needed

https://github.com/helm/helm/blob/master/docs/helm/helm_upgrade.md

rossf7:
@kopiczko Sorry for another ping but I don't see your comment. 👀 

But you mean remove the error from the key functions?

service/controller/chart/v1/key/key.go
```diff
@@ -25,6 +31,20 @@ func ConfigMapNamespace(customResource v1alpha1.Chart) string {
 	return customResource.Spec.Config.ConfigMap.Namespace
 }
 
+func HasForceUpgradeAnnotation(customResource v1alpha1.Chart) (bool, error) {
+	val, ok := customResource.Annotations[ForceHelmUpgradeAnnotationName]
+	if !ok {
+		return false, nil
+	}
+
+	result, err := strconv.ParseBool(val)
```
kopiczko: My comment was somehow lost sorry :)

I meant this when this return error? I'd like `HasForceUpgradeAnnotation` to not return error. In general we aim for `key` package functions to not return errors.

kopiczko:
```
ParseBool returns the boolean value represented by the string. It accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False. Any other value returns an error
```

Maybe it's ok to return `false` on ParseBool error? I don't know at this point. I leave the decision to you.

rossf7: Yes, I think its best to return false if ParseBool fails. As someone might add the annotation manually and fat finger it.

But I've improved the logs to also show whether force is enabled. As I think that is important for debugging.

rossf7:
Tested on `gauss` and going with this.

CI error on chartconfig controller is some permissions thing accessing github.

PR #238, opened 2019-06-05
Add change log for force upgrade support

rossf7:
Towards giantswarm/giantswarm#5745

I missed the changelog 😱. Getting it straight.


rossf7:
So cool the build is green finally. Thx! @sslavic @tomahawk28 <3

PR #242, opened 2019-06-12
vendore update: helmclient

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/6157

Upgrading helmclient only

### DIFF
--- a/helm/chart-operator-chart/templates/rbac.yaml
+++ b/helm/chart-operator-chart/templates/rbac.yaml
@@ -102,6 +102,7 @@ rules:
   verbs:
   - "create"
   - "get"
+  - "update"
 - nonResourceURLs:
   - "/"
   - "/healthz"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -221,14 +221,14 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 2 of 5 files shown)


helm/chart-operator-chart/templates/rbac.yaml
```diff
@@ -102,6 +102,7 @@ rules:
   verbs:
   - "create"
   - "get"
+  - "update"
```
tomahawk28: Adding rbac due to an error: `W 06/12 12:23:59 retrying backoff in '6.632270629s' due to error | backoff/notifier.go:13
	/go/src/github.com/giantswarm/chart-operator/vendor/github.com/giantswarm/helmclient/ensure_tiller_installed.go:261:
	deployments.extensions "tiller-deploy" is forbidden: User "system:serviceaccount:giantswarm:chart-operator" cannot update resource "deployments" in API group "extensions" in the namespace "giantswarm"`

tomahawk28:
Only need to prepare for draughtman since cluster-operator is ready.

rossf7:
OK but as this is merged we will upgrade tiller in all TCs running the `0-7-stable` release channel.

I think that is OK but please keep an eye on #alerts for problems.

PR #243, opened 2019-06-18
Vendor update: helmclient 

tomahawk28:
toward: https://github.com/giantswarm/giantswarm/issues/6157


tomahawk28:
Oops, i will update it now.

tomahawk28:
Too old, will retry after 1.14 sort it out.

PR #246, opened 2019-06-20
Add network policy

corest:
Towards https://github.com/giantswarm/giantswarm/issues/4033

### DIFF
--- a/helm/chart-operator-chart/Chart.yaml
+++ b/helm/chart-operator-chart/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for the chart-operator
 name: chart-operator-chart
-version: 0.7.1
+version: 0.8.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project's packages adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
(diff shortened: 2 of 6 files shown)


tomahawk28:
Don't you need to update helm/chart-operator-chart/templates/rbac.yaml?

corest:
@tomahawk28 nope. For what purpose?

tomahawk28:
I thought I need to list resource type in rbac if you want to create it. Maybe that's not true in the initial setup, sorry.

corest:
Network policy installed by tiller, in the same way, as for other apps. And tiller has `cluster-admin` role. Am I missing something here?

corest:
But the thing you've mentioned may apply here https://github.com/giantswarm/helmclient/pull/106 as chart-operator in control-plane needs to install tiller in tenant cluster

helm/chart-operator-chart/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for the chart-operator
 name: chart-operator-chart
-version: 0.7.1
+version: 0.8.0
```
rossf7: @corest Good to bump the versions here. If you haven't already you also need to bump the release channel in the WIP version of cluster-operator.

https://github.com/giantswarm/cluster-operator/blob/4161b4dfdcb02b73c4574d1c6f9335ac0f514587/pkg/v17/resource/chartoperator/resource.go#L27

corest: I'll bump version in cluster-operator later as there is another change expected to arrive here first

PR #247, opened 2019-06-24
tighten PSP security

glitchcrab:
towards giantswarm/giantswarm/issues/6013

chart-operator already runs as non root, so this PR just improves the existing PSP.

### DIFF
--- a/helm/chart-operator-chart/templates/psp.yaml
+++ b/helm/chart-operator-chart/templates/psp.yaml
@@ -12,12 +12,12 @@ spec:
     rule: MustRunAs
     ranges:
       - min: 1
-      - max: 65535
+        max: 65535
   fsGroup:
     rule: MustRunAs
     ranges:
       - min: 1
-      - max: 65535
+        max: 65535
   seLinux:
     rule: RunAsAny
   supplementalGroups:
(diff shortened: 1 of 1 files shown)


glitchcrab:
Not entirely sure what's going on with the CI failures here.

glitchcrab:
@giantswarm/team-batman CI tests now fixed :)

MarcelMue:
Why was there nothing added to the changelog here? I feel like this is a notable change, right?

puja108:
Hmm, for now we've been adding stuff to the changelog here that is user-facing, not stuff that is about control plane or Giant Swarm internals. As the changelogs get aggregated to the release level it might confuse users. WDYT?

MarcelMue:
This operator is running in tenants though.

rossf7:
I'm not sure if the chart-operator changelog gets propagated to the release.

IIRC in the past we've added chart-operator changelogs to cluster-operator since it deploys chart-operator in tenant clusters.

MarcelMue:
They get propagated because `chart-operator` is a component for CP and TC - on CP it's changelogs are added e.g. from here https://github.com/giantswarm/chart-operator/blob/master/service/controller/chartconfig/v6/version_bundle.go#L12

I only wanted to bring this up to spread awareness between PRs and changelogs. I think it would be nice to have some more context for the changes for our customers.

PR #251, opened 2019-07-02
Cancelling chart/release resource when chart had been cordon

tomahawk28:

Do the same practice as https://github.com/giantswarm/chart-operator/pull/250, but applied into `chart/release` resource.

### DIFF
--- a/service/controller/chart/v1/key/key.go
+++ b/service/controller/chart/v1/key/key.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/chart-operator/pkg/annotation"
 )
 
 const (
@@ -31,6 +33,14 @@ func ConfigMapNamespace(customResource v1alpha1.Chart) string {
 	return customResource.Spec.Config.ConfigMap.Namespace
 }
 
+func CordonReason(customObject v1alpha1.Chart) string {
+	return customObject.GetAnnotations()[annotation.CordonReason]
+}
+
+func CordonUntil(customObject v1alpha1.Chart) string {
+	return customObject.GetAnnotations()[annotation.CordonUntilDate]
+}
+
 func HasForceUpgradeAnnotation(customResource v1alpha1.Chart) bool {
 	val, ok := customResource.Annotations[ForceHelmUpgradeAnnotationName]
 	if !ok {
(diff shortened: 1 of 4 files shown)


service/controller/chart/v1/key/key.go
```diff
@@ -31,6 +33,14 @@ func ConfigMapNamespace(customResource v1alpha1.Chart) string {
 	return customResource.Spec.Config.ConfigMap.Namespace
 }
 
+func CordonReason(customObject v1alpha1.Chart) string {
```
rossf7:
```suggestion
func CordonReason(customResource v1alpha1.Chart) string {
```

This is the new stuff that will last a long time. So we should use the latest naming everywhere.

service/controller/chart/v1/resource/release/current.go
```diff
@@ -2,11 +2,13 @@ package release
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/giantswarm/helmclient"
 	"github.com/giantswarm/microerror"
 
 	"github.com/giantswarm/chart-operator/service/controller/chart/v1/key"
+	"github.com/giantswarm/operatorkit/controller/context/resourcecanceledcontext"
```
rossf7: Wrong block. Lets make the new stuff shiny 😉

service/controller/chart/v1/resource/release/current.go
```diff
@@ -15,6 +17,15 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
 	}
 
+	if key.IsCordoned(cr) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("release %#q has been cordoned until %#q due to reason %#q ", key.ReleaseName(cr), key.CordonUntil(cr), key.CordonReason(cr)))
+
+		resourcecanceledcontext.SetCanceled(ctx)
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+
+		return nil, nil
+	}
```
rossf7: This is all we need for this PR. But just a reminder the status resources for both controllers should set in the CR status that it is cordoned.

service/controller/chart/v1/resource/release/current.go
tomahawk28: 😱, I thought that I saw all this, sorry.

service/controller/chart/v1/resource/release/current.go
tomahawk28: Will bring it as next PR for chart-operator

service/controller/chart/v1/resource/release/current.go
rossf7: Np ❤️ I forget this too sometimes!

service/controller/chart/v1/key/key.go
```diff
@@ -47,6 +57,18 @@ func HasForceUpgradeAnnotation(customResource v1alpha1.Chart) bool {
 	return result
 }
 
+func IsCordoned(customResource v1alpha1.Chart) bool {
+	_, reasonOk := customResource.Annotations[annotation.CordonReason]
+	_, untilOk := customResource.Annotations[annotation.CordonUntilDate]
+
+	if reasonOk && untilOk {
```
kopiczko: Same as in other PR. We should check the date. It should also be reflected in tests.

PR #256, opened 2019-07-09
Remove helper function

tomahawk28:
Idea from @xh3b4sd 
https://github.com/giantswarm/chart-operator/pull/221#discussion_r301210172

- Remove `helper` usage since it's adding an additional layer to make it fully understand our collector functionality. It's unnecessary.


service/collector/chart_resource.go
```diff
@@ -67,21 +64,21 @@ func NewChartResource(config ChartResourceConfig) (*ChartResource, error) {
...
 		ch <- prometheus.MustNewConstMetric(
 			chartConfigDesc,
 			prometheus.GaugeValue,
 			gaugeValue,
-			chartConfig.chartName,
-			chartConfig.channelName,
-			chartConfig.releaseName,
-			chartConfig.releaseStatus,
-			chartConfig.namespace,
+			key.ChartName(chartConfig),
```
tomahawk28: Replace all assigning variables using `key`

PR #257, opened 2019-07-09
following logging formats as we agree on fmt.

tomahawk28:
From a conversation with @xh3b4sd, I found several log messages still not following our fmt. 
https://github.com/giantswarm/fmt/blob/master/go/log_messages.md


tomahawk28:
It's mostly changes from `%s` to `%#v`.

PR #260, opened 2019-07-11
Chart cordon metric

tomahawk28:

This is new cordon metric support for `chart` CR.

### DIFF
--- a/service/collector/chart_resource.go
+++ b/service/collector/chart_resource.go
@@ -1,0 +1,92 @@
+package collector
+
+import (
+	"fmt"
+
+	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/prometheus/client_golang/prometheus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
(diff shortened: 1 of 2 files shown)


service/collector/chart_resource.go
```diff
@@ -0,0 +1,92 @@
...
+		return microerror.Mask(err)
+	}
+
+	for _, chart := range charts.Items {
+
+		if !key.IsCordoned(chart) {
+			continue
+		}
+
+		t, err := convertToTime(key.CordonUntil(chart))
```
tomahawk28: Be aware that I'm using the helper function in other files :)

service/collector/chart_resource.go
```diff
@@ -0,0 +1,92 @@
...
+		if err != nil {
+			c.logger.Log("level", "warning", "message", "could not convert cordon-until", "stack", fmt.Sprintf("%#v", err))
+			continue
+		}
+
+		ch <- prometheus.MustNewConstMetric(
+			chartCordonExpireTimeDesc,
+			prometheus.GaugeValue,
+			float64(t.Unix()),
+			key.ReleaseName(chart),
```
tomahawk28: This is `ReleaseName` but actually link into `.spec.Name`

tomahawk28:
Damn, I will move this part of functionality to app-operator soon.

PR #263, opened 2019-07-16
Cordon status for chart CR

tomahawk28:

cordon status update for `chart` CR.

### DIFF
--- a/service/controller/chart/v1/resource/status/create.go
+++ b/service/controller/chart/v1/resource/status/create.go
@@ -42,11 +42,25 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		return microerror.Mask(err)
 	}
 
+	var status, reason string
+	{
+		if key.IsCordoned(cr) {
+			status = releaseStatusCordoned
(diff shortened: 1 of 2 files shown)


service/controller/chart/v1/resource/status/create.go
```diff
@@ -62,17 +76,6 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 
 		currentCR.Status = desiredStatus
 
-		if releaseContent.Status != releaseStatusDeployed {
```
tomahawk28: We don't need check status twice.

service/controller/chart/v1/resource/status/resource.go
```diff
@@ -62,6 +63,9 @@ func equals(a, b v1alpha1.ChartStatus) bool {
 	if a.Release.LastDeployed != b.Release.LastDeployed {
 		return false
 	}
+	if a.Reason != b.Reason {
```
tomahawk28: There are cases chart failed multiple times, so we need to compare the `reason` as well.

PR #265, opened 2019-07-25
Exposes version

TheoBrigitte:
This PR exposes the version of chart-operator via version command and version endpoint.
Also moves project information under pkg/project.

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -1,0 +1,29 @@
+package project
+
+var (
+	description = "The chart-operator deploys Helm charts by reconciling against a CNR repository."
+	name        = "chart-operator"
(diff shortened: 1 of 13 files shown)


pkg/project/project.go
```diff
@@ -0,0 +1,29 @@
+package project
+
+var (
+	description = "The chart-operator deploys Helm charts by reconciling against a CNR repository."
```
rossf7:
```suggestion
	description = "chart-operator is our agent for deploying Helm charts as releases."
```

Nitpick but this is out of date. Only the chartconfig controller uses CNR and its deprecated. The chart controller uses Helm repositories.

PR #267, opened 2019-07-30
Update workflow to push into giantswarm catalog

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/6654

Adding chart-operator into https://github.com/giantswarm/giantswarm-catalog.


.circleci/config.yml
```diff
@@ -213,3 +216,14 @@ workflows:
               only: master
           requires:
           - deploy
+  push-to-app-catalog:
```
tomahawk28: Perhaps we can think different on the naming of workflow?

.circleci/config.yml
```diff
@@ -26,7 +26,10 @@ e2eTest: &e2eTest
         command: ./e2e-harness teardown
         when: always
 
-version: 2
+version: 2.1
```
tomahawk28: Not sure the best place put orbs, and versions.

.circleci/config.yml
```diff
@@ -213,3 +216,14 @@ workflows:
...
+    jobs:
+      - architect/push-to-app-catalog:
+          name: "package and push chart-operator-chart"
+          app_catalog: "giantswarm-catalog"
+          app_catalog_test: "giantswarm-test-catalog"
+          chart: "chart-operator-chart"
+          # Trigger job on git tag.
+          filters:
+            tags:
+              only: /^v.*/
```
kopiczko: You will also need to add push-to-docker: https://github.com/giantswarm/aws-operator/pull/1808/files

There is not yet merged guide we are working on: https://github.com/giantswarm/giantswarm/pull/6656/files

tomahawk28: @kopiczko Should I add `push-to-docker` as in the same workflow `push-to-catalog`?

kopiczko: Yes https://github.com/giantswarm/aws-operator/blob/15592ef325d5287fac2f4a791517da5f91301589/.circleci/config.yml#L233-L253

.circleci/config.yml
kopiczko: I think it's fine

.circleci/config.yml
```diff
@@ -213,3 +216,26 @@ workflows:
               only: master
           requires:
           - deploy
+  push-to-app-catalog:
+    jobs:
+      - architect/push-to-docker:
```
kopiczko:
```suggestion
      - architect/push-to-docker:
        name: "build and push chart-operator docker image"
```

.circleci/config.yml
```diff
@@ -213,3 +216,26 @@ workflows:
...
+          image: "quay.io/giantswarm/chart-operator"
+          username_envar: "QUAY_USERNAME"
+          password_envar: "QUAY_PASSWORD"
+          requires:
+            - build
+          filters:
+            tags:
+              only: /^v.*/
+      - architect/push-to-app-catalog:
+          name: "package and push chart-operator-chart"
```
kopiczko:
```suggestion
          name: "package and push chart-operator helm chart"
```

.circleci/config.yml
```diff
@@ -213,3 +216,26 @@ workflows:
...
+          app_catalog: "giantswarm-catalog"
+          app_catalog_test: "giantswarm-test-catalog"
+          chart: "chart-operator-chart"
```
kopiczko: You'll need to prepare a new chart. See https://gigantic.slack.com/archives/C04TGHDEF/p1559728459002200

```suggestion
          chart: "chart-operator"
```

kopiczko:
I also proposed some changes to align in aws-operator https://github.com/giantswarm/aws-operator/pull/1816

.circleci/config.yml
```diff
@@ -155,10 +156,10 @@ workflows:
               only: /^v.*/
 
       - architect/push-to-app-catalog:
-          name: "package and push chart-operator-chart"
+          name: "package and push chart-operator helm chart"
           app_catalog: "giantswarm-catalog"
           app_catalog_test: "giantswarm-test-catalog"
-          chart: "chart-operator-chart"
+          chart: "chart-operator"
           # Make sure docker image is successfully built.
```
kopiczko: I'd like to remove this comment. This is obvious.

tomahawk28:
@kopiczko Thanks for your input, I think `aws-operator` also have an issue with a name inside a file. 
```
Job 'build and push aws-operator helm chart' requires 'architect/push-to-docker', which is the name of 0 other jobs in workflow 'build'
```
I think it's happening new variable, `name`. 
https://circleci.com/gh/giantswarm/aws-operator/21708?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

.circleci/config.yml
rossf7: @tomahawk28 I don't see the 2nd chart in the PR. Don't we need this?

tomahawk28: ah, this is the reason why I failed as below. 
```
mkdir build && helm package ./helm/chart-operator --destination ./build 00:00
Exit code: 1
#!/bin/bash -eo pipefail
mkdir build && helm package ./helm/chart-operator --destination ./build
Error: chart metadata (Chart.yaml) missing
Exited with code 1
```

tomahawk28: Damm, it blocked by gitignore rule that we put on 😆 
```
  4 # build result
  5 chart-operator
  6 

```

helm/chart-operator/values.yaml
```diff
@@ -0,0 +1,33 @@
+# Default values for chart-operator-chart.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
```
kopiczko: Let's please align this file with https://github.com/giantswarm/aws-operator/blob/28a55852c391fce54ca13629aa99caca1fe12824/helm/aws-operator/values.yaml

tomahawk28:
@kopiczko Do you have similar issue like below?
```
cnr: error: 405 Client Error: METHOD NOT ALLOWED for url: https://quay.io/cnr/api/v1/packages/giantswarm/chart-operator?force=true
```

Anything I need to update other than values.yaml?

helm/chart-operator-chart/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for the chart-operator
-name: chart-operator-chart
+name: chart-operator
```
kopiczko: chart-operator-chart should stay as it is.

helm/chart-operator/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+description: A Helm chart for the chart-operator
+name: chart-operator-chart
```
kopiczko: This should be
```suggestion
name: chart-operator
```

tomahawk28:
@kopiczko Have applied same things in aws-operator, PTAL.

.circleci/config.yml
```diff
@@ -137,10 +140,31 @@ jobs:
 
 workflows:
   version: 2
-  build_e2e:
+  build:
     jobs:
       - build
 
+      - architect/push-to-docker:
```
kopiczko: please add `name:` https://github.com/giantswarm/aws-operator/blob/587cc78eee55393d1748c3715f5860f364dad0b0/.circleci/config.yml#L234

.circleci/config.yml
```diff
@@ -137,10 +140,31 @@ jobs:
...
+          image: "quay.io/giantswarm/chart-operator"
+          username_envar: "QUAY_USERNAME"
+          password_envar: "QUAY_PASSWORD"
+          requires:
+            - build
+          filters:
```
kopiczko:
```suggestion
          # Needed to trigger job also on git tag.
          filters:
```

.circleci/config.yml
```diff
@@ -137,10 +140,31 @@ jobs:
...
+            tags:
+              only: /^v.*/
+
+      - architect/push-to-app-catalog:
+          app_catalog: "giantswarm-catalog"
+          app_catalog_test: "giantswarm-test-catalog"
+          chart: "chart-operator"
+          requires:
+            - architect/push-to-docker
+          # Trigger job on git tag.
```
kopiczko:
```suggestion
          # Needed to trigger job also on git tag.
```

helm/chart-operator/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+description: A Helm chart for the chart-operator
+name: chart-operator
+version: 0.9.0
```
kopiczko:
```suggestion
version: [[ .Version ]]
```

helm/chart-operator/values.yaml
```diff
@@ -0,0 +1,35 @@
...
+# room for those suffixes.
+#
+# NOTE: All values under resource key need to be used with `tpl` to render them
+# correctly in the templates. This is because helm doesn't template values.yaml
+# file and it has to be a valid json. Example usage:
+#
+#     {{ tpl .Values.resource.default.name . }}.
+#
+resource:
+  clusterDNSIP: 172.31.0.10
```
kopiczko: I think this should go out of `resource`. I'd like to keep here only resource names. This doesn't also have to be wrapped with `tpl`.

rossf7: Agree with @kopiczko but also we need to be very careful here because `clusterDNSIP` is set by cluster-operator when it installs chart-operator.

IIRC this is so chart-operator can install CoreDNS in new clusters. If we change the values structure this will probably fail.

https://github.com/giantswarm/cluster-operator/blob/9e12a2aedf6647c4aeaa6ef84b4fb0326205e199/pkg/v18/resource/chartoperator/desired.go#L25

tomahawk28: @rossf7 But that resource using `chart-operator-chart` helm chart so we are okay to move it.

helm/chart-operator/Chart.yaml
rossf7: Thanks @kopiczko yes since this is the chart for the app catalog we should use this.

helm/chart-operator/templates/00-service-account.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
rossf7: @tomahawk28 Why did we start using this `00-service-account.yaml` naming?  We don't use this now so why do we need it?

helm/chart-operator/values.yaml
rossf7: > I think this should go out of resource. I'd like to keep here only resource names. 

Like Pawel says this block is for the resource naming and `clusterDNSIP` isn't part of that. So I'd like us to keep the original structure and make this a top level value.

rossf7: In the chartoperator resource in app-operator we will need to set this value. Like we do now in cluster-operator.

https://github.com/giantswarm/cluster-operator/blob/9e12a2aedf6647c4aeaa6ef84b4fb0326205e199/pkg/v18/resource/chartoperator/desired.go#L25

helm/chart-operator/templates/00-service-account.yaml
tomahawk28: @rossf7 I was following the naming from new helm chart in aws-operator, and it's easier to figure out the order of installing components. WDYT?

tomahawk28: @rossf7 I deleted all the numbers :) PTAL!

helm/chart-operator/README.md
```diff
@@ -0,0 +1,14 @@
...
+|--------------------|---------------------------------------------|-------------------------------------------------------------------|
+| `name`             | The name of the operator                    | `chart-operator`                                                  |
+| `namespace`        | The namespaces of the operator              | `giantswarm`                                                      |
+| `port`             | The port of the operator container          | `8000`                                                            |
+| `image.registry`   | The operator container image registry       | `quay.io`                                                         |
+| `image.repository` | The operator container image repository     | `giantswarm/chart-operator`                                       |
+| `image.tag`        | The operator container image tag            | `[latest commit SHA]`                                             |
+| `resources`        | The operator pod resource requests & limits | `request: cpu:100m, memory:75Mi;  limits: cpu:250m, memory:250Mi` |
```
rossf7: Not blocking but I don't think this README is necessary. We have lots of helm charts and this is standard operator stuff. 

IMO docs like this don't add much value and get stale quickly.

helm/chart-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+            path: config.yaml
+      priorityClassName: giantswarm-critical
+      serviceAccountName: {{ tpl .Values.resource.default.name  . }}
+      tolerations:
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
+      dnsPolicy: None
+      dnsConfig:
+        nameservers:
+        - {{ .Values.clusterDNSIP }}
```
rossf7: <3

helm/chart-operator/templates/rbac.yaml
```diff
@@ -0,0 +1,165 @@
...
+    app: {{ .Values.project.name }}
+    giantswarm.io/service-type: "managed"
+subjects:
+  - kind: ServiceAccount
+    name: {{ tpl .Values.resource.default.name  . }}
+    namespace: {{ tpl .Values.resource.default.namespace . }}
+roleRef:
+  kind: ClusterRole
+  name: {{ tpl .Values.resource.psp.name . }}
+  apiGroup: rbac.authorization.k8s.io
```
rossf7: Line ending is wrong here.

PR #271, opened 2019-08-10
Improve unit tests for release resource

rossf7:
Aligning with https://github.com/giantswarm/fmt/blob/master/go/unit_tests.md.


rossf7:
Build is green. Unit tests only. Going with this.

PR #273, opened 2019-08-11
Use patch for annotations not update

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/6696

The `values-md5-checksum` annotation is now set via a patch operation. This is safer because both chart-operator and app-operator need to set annotations.

See giantswarm/giantswarm#6696 (comment).

### DIFF
--- a/service/controller/chart/v1/resource/release/resource.go
+++ b/service/controller/chart/v1/resource/release/resource.go
@@ -2,6 +2,7 @@ package release
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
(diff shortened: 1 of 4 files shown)


service/controller/chart/v1/resource/release/resource.go
```diff
@@ -111,22 +111,32 @@ func (r *Resource) updateAnnotations(ctx context.Context, cr v1alpha1.Chart, rel
 	currentChecksum := key.ValuesMD5ChecksumAnnotation(*currentCR)
 
 	if releaseState.ValuesMD5Checksum != currentChecksum {
-		if currentCR.Annotations != nil {
-			annotations = currentCR.Annotations
+		annotations := map[string]string{
+			key.ValuesMD5ChecksumAnnotationName: releaseState.ValuesMD5Checksum,
+		}
+
+		patches := []Patch{
```
tomahawk28: @giantswarm/team-batman This is wrong, it would replace the whole annotation (checked on ginger CP), not just `md5sum`, to do it correctly we need to use `add` operation.

PR #280, opened 2019-09-05
Update vendor to use latest operatorkit

rossf7:


helm/chart-operator-chart/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: A Helm chart for the chart-operator
 name: chart-operator-chart
-version: 0.9.0
+version: 0.10.0
```
rossf7: Bumping the chart version to be extra paranoid.

PR #281, opened 2019-09-06
Annotation for deleting chartconfig CRs

rossf7:
Towards giantswarm/giantswarm#6800

Adds support for the `chart-operator.giantswarm.io/delete-custom-resource-only` annotation. 

This is added by cluster-operator once the migration from chartconfig to app CR is complete. The chartconfig CR is then deleted but the helm release is not.


service/controller/chartconfig/v7/resource/chart/current.go
```diff
@@ -17,7 +18,11 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
-	if key.IsCordoned(customObject) {
+	// Delete annotation is set when the chartconfig CR has been migrated to an
+	// app CR and can be safely deleted.
+	_, hasDeleteAnnotation := customObject.Annotations[annotation.DeleteCustomResourceOnly]
```
kopiczko: Shouldn't we have a function in the key package as for cordon annotation?

pkg/annotation/annotation.go
```diff
@@ -10,4 +10,8 @@ const (
 	// CordonUntilDate is the name of the annotation that indicates
 	// the expiration date of rule of this cordon.
 	CordonUntilDate = "chart-operator.giantswarm.io/cordon-until"
+
+	// DeleteCustomResourceOnly is the name of the annotation that indicates
+	// the custom resource should be deleted without deleting the Helm release.
+	DeleteCustomResourceOnly = "chart-operator.giantswarm.io/delete-custom-resource-only"
```
kopiczko: Is the value ignored?

service/controller/chartconfig/v7/resource/chart/current.go
```diff
@@ -17,7 +18,11 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
+
+	if key.IsCordoned(customObject) && !hasDeleteAnnotation {
```
tomahawk28: I don't get this part, why we don't simply use it `Cordon` annotation? 
I also look into `service/controller/chartconfig/v7/resource/chart/delete.go` but it only return nil when it has`deleteAnnotation`. Perhaps you want to keep update & create CRUD features as on?

pkg/annotation/annotation.go
tomahawk28: I think we don't need to care of value in here.

service/controller/chartconfig/v7/resource/chart/current.go
rossf7: Yes good point. I'll add that.

pkg/annotation/annotation.go
rossf7: Yes it's just the presence of the annotation that matters.

service/controller/chartconfig/v7/resource/chart/current.go
rossf7: We cordon the app at the start of the migration. Once the app status is deployed we add this annotation.

This indicates the migration is complete and we delete the chartconfig CR. Here is the code in cluster-operator.

https://github.com/giantswarm/cluster-operator/blob/f9ca993418f053d705dade4d2cf72976cee40f4d/pkg/v20/resource/appmigration/create.go#L117-L138

pkg/annotation/annotation.go
rossf7: In the key helper it now checks the value is true.

rossf7: I missed using the key func in the current state but this is now fixed. It does now check the value is true and I agree that is better.

PR #283, opened 2019-09-11
Fix templating for RBAC

rossf7:
Fixes the templating for my previous change.


rossf7:
Tested on gauss.

tomahawk28:
@sslavic `chart-operator` is for making a catalog, `chart-operator-chart` is for quay uploading.

PR #287, opened 2019-09-17
Fix chart-operator chart templating

rossf7:
The templating in the App Catalog chart was incorrect. The resource values are just for ensuing multiple instances of the operator can be installed.

Aligns with the appr chart. https://github.com/giantswarm/chart-operator/blob/5440b976e74d41bc06fd6a133356bd2dfae214ea/helm/chart-operator-chart/values.yaml


rossf7:
Tested on `ghost`.

PR #292, opened 2019-09-24
Add init container to wait for quay.io to be resolvable

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/6800

- Adds an init container which waits for quay.io to be resolvable using 8.8.8.8.
- This is because chart-operator has its own DNS settings so it can install coredns in new tenant clusters.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -179,7 +179,7 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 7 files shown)


rossf7:
CI for chartconfig CRs fails because its trying to get the release index from installations.

rossf7:
Tested on `ginger` going with this.

PR #297, opened 2019-09-30
Don't update chart-operator chart CR in case of failure

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/7095

Canceling a release resource in case of it's chart-operator release since it would re-installed by app-operator.

### DIFF
--- a/service/controller/chart/v1/resource/release/create_test.go
+++ b/service/controller/chart/v1/resource/release/create_test.go
@@ -75,6 +75,8 @@ func Test_Resource_Release_newCreate(t *testing.T) {
 			HelmClient: helmclienttest.New(helmclienttest.Config{}),
 			K8sClient:  k8sfake.NewSimpleClientset(),
 			Logger:     microloggertest.New(),
+
+			ProjectName: "chart-operator",
 		}
 
 		newResource, err = New(c)
--- a/service/controller/chart/v1/resource/release/current.go
+++ b/service/controller/chart/v1/resource/release/current.go
@@ -36,7 +36,7 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	}
 
 	if releaseContent.Status == "FAILED" && releaseContent.Name == r.projectName {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("not updating release %#q since it's bootstrap from app-operator", releaseContent.Name))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("not updating own release %#q since it's %#q", releaseContent.Name, releaseContent.Status))
 
 		resourcecanceledcontext.SetCanceled(ctx)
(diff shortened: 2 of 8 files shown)


service/controller/chart/v1/resource/release/current.go
```diff
@@ -35,6 +35,15 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
+	if releaseContent.Status == "FAILED" && releaseContent.Name == "chart-operator" {
```
tomahawk28: Not sure this is the right approach, we can cancel this in current.go so could save a time for other resources.

service/controller/chart/v1/resource/release/current.go
```diff
@@ -35,6 +35,15 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("not updating release %#q since it's bootstrap from app-operator", releaseContent.Name))
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("not updating own release %#q since it's %#q", releaseContent.Name, releaseContent.Status))
```

service/controller/chart/v1/resource/release/current.go
rossf7:
```suggestion
	if releaseContent.Status == "FAILED" && releaseContent.Name == r.projectName {
```

We can remove the hardcoding by using the project name. It just needs adding to the resource wiring.

rossf7: I think this is fine. Current state is called first and if the release is failed we don't want to continue.

service/controller/chart/v1/resource/release/current.go
rossf7: I'd still like the log message to be updated. If its the CP chart-operator is still managed by draughtsman not app-operator.

tomahawk28: I missed this part, sorry.

service/controller/chart/v1/resource/release/create_test.go
```diff
@@ -75,6 +75,8 @@ func Test_Resource_Release_newCreate(t *testing.T) {
 			HelmClient: helmclienttest.New(helmclienttest.Config{}),
 			K8sClient:  k8sfake.NewSimpleClientset(),
 			Logger:     microloggertest.New(),
+
+			ProjectName: "chart-operator",
```
kopiczko: It would be good to introduce `pkg/project` here https://github.com/giantswarm/aws-operator/tree/master/pkg/project. It is automatically populated by architect.

rossf7: Good point. I did that in some other repos like pv-cleanup-operator but missed it here. I'll get it sorted.

PR #302, opened 2019-10-13
Support multiple external DNS servers

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/6874

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
-FROM quay.io/giantswarm/alpine:3.9-giantswarm
+FROM quay.io/giantswarm/alpine:3.10-giantswarm
 
 USER root
 
-RUN apk add --no-cache ca-certificates
+# bind-tools is required by the init container to use dig.
+RUN apk add --no-cache ca-certificates bind-tools
 
 USER giantswarm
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,8 +1,9 @@
-FROM quay.io/giantswarm/alpine:3.9-giantswarm
+FROM quay.io/giantswarm/alpine:3.10-giantswarm
 
 USER root
 
-RUN apk add --no-cache ca-certificates
+# bind-tools is required by the init container to use dig.
+RUN apk add --no-cache ca-certificates bind-tools
```
rossf7: Using dig instead of nslookup because it supports setting multiple name servers. Rather than creating an `alpine-dig` image I'm just adding it to the current image.

helm/chart-operator/templates/deployment.yaml
```diff
@@ -58,17 +60,17 @@ spec:
           value: "5"
       initContainers:
       - name: wait-for-quay
-        image: quay.io/giantswarm/alpine:3.10-giantswarm
+        image: "{{ .Values.image.registry }}/{{ .Values.image.name }}:{{ .Values.image.tag }}"
```
rossf7: Also fixing the retagging.

helm/chart-operator/values.yaml
```diff
@@ -7,7 +7,7 @@ externalDNSIP: 8.8.8.8
 
 image:
   registry: "quay.io"
-  repository: "giantswarm/chart-operator"
+  name: "giantswarm/chart-operator"
```
rossf7: Aligning with fmt. https://github.com/giantswarm/fmt/blob/master/helm/image_templating.md

PR #308, opened 2019-10-24
Cancelling the resource in case of tarball error

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/7359


service/controller/chart/v1/resource/release/desired.go
```diff
@@ -27,7 +28,12 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 
 	tarballPath, err := r.helmClient.PullChartTarball(ctx, tarballURL)
 	if err != nil {
-		return nil, microerror.Mask(err)
+		r.logger.LogCtx(ctx, "level", "error", "message", "pulling chart failed", "stack", microerror.Stack(err))
```
tomahawk28: Putting a stack in the end since network error could vary.

xh3b4sd: When we expect an error the logging should only be `warning`.

service/controller/chart/v1/resource/release/desired.go
```diff
@@ -27,7 +28,12 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 
 	tarballPath, err := r.helmClient.PullChartTarball(ctx, tarballURL)
 	if err != nil {
```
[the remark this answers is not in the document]
tomahawk28: In this case, it's `executionFailedError` in `helmclient`. This is one of the network error and the fact that we can't use the matcher against `executionFailedError`, Maybe I should return `pullChartFailedError` as a customized one in `helmclient`, WDYT?
https://github.com/giantswarm/helmclient/blob/46206f66eaf4a9a4aa526457ae5d8e0124ec3d90/pull_chart_tarball.go#L39-L41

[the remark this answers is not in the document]
xh3b4sd: 100%, let's get more fine grained error handling here down the line.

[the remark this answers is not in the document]
rossf7: Yes I think returning pullChartFailedError is good.

tomahawk28:
@giantswarm/team-batman @giantswarm/sig-operator It's ready, PTAL

service/controller/chart/v1/resource/release/create.go
```diff
@@ -29,7 +29,12 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 
 		tarballPath, err := r.helmClient.PullChartTarball(ctx, tarballURL)
 		if err != nil {
```
xh3b4sd: Why is this still matching all errors? We said this does not make sense.

tomahawk28: I didn't see there are other two usage 🤦‍♂ Sorry.

PR #310, opened 2019-10-30
Remove usage of e2e-harness framework package in e2e tests

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/4606

Same change Mr @kopiczko  made in https://github.com/giantswarm/azure-operator/pull/583 to remove usage of the framework package.


rossf7:
Yes these tests are just needed while we have the chartconfig controller. Once it's gone we can drop them.

PR #313, opened 2019-11-19
Cancel reconciliation when there are too many tiller pods

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7603

### DIFF
--- a/service/controller/chart/v1/resource/tiller/resource.go
+++ b/service/controller/chart/v1/resource/tiller/resource.go
@@ -58,7 +58,7 @@ func (r *Resource) ensureTillerInstalled(ctx context.Context) error {
 	}
 	err := r.helmClient.EnsureTillerInstalledWithValues(ctx, values)
 	if helmclient.IsTooManyResults(err) {
-		r.logger.LogCtx(ctx, "level", "debug", "message", "currently too many tiller pods")
+		r.logger.LogCtx(ctx, "level", "debug", "message", "currently too many tiller pods due to upgrade")
 		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation")
 		reconciliationcanceledcontext.SetCanceled(ctx)
 		return nil
--- a/service/controller/chart/v1/resource/tiller/resource.go
+++ b/service/controller/chart/v1/resource/tiller/resource.go
@@ -6,6 +6,7 @@ import (
 	"github.com/giantswarm/helmclient"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
+	"github.com/giantswarm/operatorkit/controller/context/reconciliationcanceledcontext"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


service/controller/chart/v1/resource/tiller/resource.go
```diff
@@ -56,7 +57,12 @@ func (r *Resource) ensureTillerInstalled(ctx context.Context) error {
 		"spec.template.spec.tolerations[0].operator=Exists",
 	}
 	err := r.helmClient.EnsureTillerInstalledWithValues(ctx, values)
-	if err != nil {
+	if helmclient.IsTooManyResults(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "currently too many tiller pods")
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation")
+		reconciliationcanceledcontext.SetCanceled(ctx)
+		return nil
```
rossf7: During upgrades there can be multiple pods.

chart-operator can't do anything without a tiller pod. So it's safer to cancel the reconciliation and wait for the tiller deployment.

service/controller/chart/v1/resource/tiller/resource.go
```diff
@@ -56,7 +57,12 @@ func (r *Resource) ensureTillerInstalled(ctx context.Context) error {
 		"spec.template.spec.tolerations[0].operator=Exists",
 	}
 	err := r.helmClient.EnsureTillerInstalledWithValues(ctx, values)
-	if err != nil {
+	if helmclient.IsTooManyResults(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "currently too many tiller pods")
```
kopiczko: Does it make sense to be more precise here?

```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", "currently too many tiller pods due to upgrade")
```

rossf7: Good shout. Done.

service/controller/chart/v1/resource/tiller/resource.go
MarcelMue: This seems like a dangerous situation to me. 
Failure case: There is another tiller running in the ns but no upgrade is happening (e.g. customer installed it).

There is no way to detect this situation anymore or alert on it. Is there a too many tillers metric? Can we add one and alert if there is more than 1 tiller in the namespace for e.g. 1 hour?

rossf7: Good point. I'll add that metric to the collector here since chart-operator is deployed everywhere.

Of course I'd rather invest the time in the Helm 3 upgrade but we are where we are.

MarcelMue: I agree with the helm 3 point but I can't tell how far away we are from that / how much effort that's going to be.

rossf7: We're doing the [Helm 2.16.1 upgrade](https://github.com/giantswarm/giantswarm/issues/7675) first for K8s 1.16. We can then plan the [Helm 3 upgrade](https://github.com/giantswarm/giantswarm/issues/7676).

I created issues yesterday but no ETA on Helm 3 yet.

rossf7:
Tested on `ghost`. New metric is working.

```
$ kg port-forward chart-operator-65d666656c-d298b 8000:8000

$ curl -s http://localhost:8000/metrics | grep chart_operator_tiller_running_pods
# HELP chart_operator_tiller_running_pods Tiller running pods.
# TYPE chart_operator_tiller_running_pods gauge
chart_operator_tiller_running_pods{namespace="kube-system"} 1
```

stone-z:
@rossf7 is this ready to merge? If so I'll include it in the kubernetes 1.16 release

PR #314, opened 2019-11-21
Vendor dependencies, update k8s to 1.16.3

sslavic:
Towards giantswarm/giantswarm#7675


service/controller/chart/chart.go
```diff
@@ -4,9 +4,9 @@ import (
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
 	"github.com/giantswarm/helmclient"
-	"github.com/giantswarm/k8sclient/k8scrdclient"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
+	"github.com/giantswarm/operatorkit/client/k8scrdclient"
```
sslavic: @tomahawk28 I'm curious what was the reason to revert this change in `chart-operator`?

`client` package in `operatorkit` has been deprecated and replaced by one in `k8sclient`, see:
- https://github.com/giantswarm/operatorkit/pull/339
- https://github.com/giantswarm/k8sclient/pull/17

This actually helped us drop operatorkit dependency in multiple projects and break circular dependency.

If `k8scrdclient` in `k8sclient` is not working, we should let Tim know and potentially revert changes in multiple other projects. If it's working, we should consistently use latest one, reduce use of deprecated API to allow it to be removed.

tomahawk28: @sslavic Thanks for the input, I got errors from build step which complaining about type mismatch. 
I would try to solve this on chart-operator side first, and if not works then I would let Tim know about this.

tomahawk28:
It's not only vendor change, also about some structure changes so I would like to get the second opinion.

PR #315, opened 2019-11-25
Cancel resource if helm release does not exist

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7355

Fixes errors seen when investigating aws-operator not being deployed to ginger.

### DIFF
--- a/service/controller/chart/v1/resource/release/create.go
+++ b/service/controller/chart/v1/resource/release/create.go
@@ -53,7 +53,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		if err != nil {
 			releaseContent, err := r.helmClient.GetReleaseContent(ctx, releaseState.Name)
 			if helmclient.IsReleaseNotFound(err) {
-				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("helm release not found %#q", releaseContent.Name))
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("helm release %#q not found", releaseContent.Name))
 				r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
 				resourcecanceledcontext.SetCanceled(ctx)
 				return nil
--- a/service/controller/chart/v1/resource/release/create.go
+++ b/service/controller/chart/v1/resource/release/create.go
@@ -52,15 +52,18 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		err = r.helmClient.InstallReleaseFromTarball(ctx, tarballPath, ns, helm.ReleaseName(releaseState.Name), helm.ValueOverrides(releaseState.ValuesYAML))
 		if err != nil {
 			releaseContent, err := r.helmClient.GetReleaseContent(ctx, releaseState.Name)
-			if err != nil {
+			if helmclient.IsReleaseNotFound(err) {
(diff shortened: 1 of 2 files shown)


service/controller/chart/v1/resource/release/create.go
```diff
@@ -52,15 +52,18 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		err = r.helmClient.InstallReleaseFromTarball(ctx, tarballPath, ns, helm.ReleaseName(releaseState.Name), helm.ValueOverrides(releaseState.ValuesYAML))
 		if err != nil {
 			releaseContent, err := r.helmClient.GetReleaseContent(ctx, releaseState.Name)
-			if err != nil {
+			if helmclient.IsReleaseNotFound(err) {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("helm release not found %#q", releaseContent.Name))
+				r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+				resourcecanceledcontext.SetCanceled(ctx)
+				return nil
```
rossf7: If there is no helm release we should cancel the resource rather than triggering the operatorkit errors alert.

We have alerting for apps not being installed and it is then clearer what the problem is.

service/controller/chart/v1/resource/release/update.go
```diff
@@ -55,15 +55,18 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
 			helm.UpgradeForce(upgradeForce))
 		if err != nil {
 			releaseContent, err := r.helmClient.GetReleaseContent(ctx, releaseState.Name)
-			if err != nil {
+			if helmclient.IsReleaseNotFound(err) {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("helm release not found %#q", releaseContent.Name))
+				r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+				resourcecanceledcontext.SetCanceled(ctx)
+				return nil
+			} else if err != nil {
```
rossf7: Same fix for updating releases.

service/controller/chart/v1/resource/release/create.go
```diff
@@ -52,15 +52,18 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		err = r.helmClient.InstallReleaseFromTarball(ctx, tarballPath, ns, helm.ReleaseName(releaseState.Name), helm.ValueOverrides(releaseState.ValuesYAML))
 		if err != nil {
 			releaseContent, err := r.helmClient.GetReleaseContent(ctx, releaseState.Name)
-			if err != nil {
+			if helmclient.IsReleaseNotFound(err) {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("helm release not found %#q", releaseContent.Name))
```
tomahawk28:
```suggestion
				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("helm release %#q not found ", releaseContent.Name))
```

rossf7: That's nicer! Done.

rossf7:
> I don't get how we would end up with no release, is that means it failed for helm chart specific reasons?

When we call `InstallReleaseFromTarball` or `UpdateReleaseFromTarball` if there is an error we try get the release details to see if it has status failed.

But if the error prevented the creation of the release we didn't handle that. This is a fix for that.

PR #318, opened 2019-12-02
Make registry configurable for tiller image

rossf7:
Towards giantswarm/giantswarm#7773

Making the image registry configurable for the Tiller resource for AWS China.


main.go
```diff
@@ -114,6 +114,7 @@ func mainWithError() error {
 	daemonCommand.PersistentFlags().String(f.Service.CNR.Address, "https://quay.io", "Address used to connect to CNR, defaults to quay's managed offering.")
```
sslavic: @rossf7 is this duplicate now i.e. can it be removed?

rossf7: It's used for pulling tarballs from Appr. We need both because in China we pull images from Aliyun but tarballs are still from Quay.

PR #319, opened 2019-12-02
Update project wiring

rossf7:
chart-operator was behind because the wiring doesn't use the `pkg/project` package.

Updating it to get on track and make later refactoring easier.

### DIFF
--- a/main.go
+++ b/main.go
@@ -52,15 +52,10 @@ func mainWithError() error {
 		var newService *service.Service
 		{
 			c := service.Config{
-				Flag:   f,
 				Logger: newLogger,
-				Viper:  v,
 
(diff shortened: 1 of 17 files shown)


main.go
```diff
@@ -52,15 +52,10 @@ func mainWithError() error {
...
 				Logger: newLogger,
-				Viper:  v,
 
-				Description: project.Description(),
-				GitCommit:   project.GitSHA(),
-				ProjectName: project.Name(),
-				Source:      project.Source(),
-				Version:     project.Version(),
+				Flag:  f,
+				Viper: v,
```
rossf7: Wiring is simplified and we get values from pkg/project when needed.

PR #321, opened 2019-12-02
Vendor operatorkit updated controller internals

rossf7:
Uses the updated operatorkit internals from https://github.com/giantswarm/operatorkit/pull/338.

I think its good to test these in multiple operators since the changes under the hood are quite big.


Gopkg.toml
```diff
@@ -93,7 +93,7 @@ required = [
   name = "github.com/giantswarm/micrologger"
 
 [[constraint]]
-  branch = "master"
+  branch = "cluster-controller"
```

service/controller/chartconfig/chart_config.go
```diff
@@ -44,43 +39,24 @@ func NewChartConfig(config Config) (*ChartConfig, error) {
...
 	}
-	if config.K8sExtClient == nil {
-		return nil, microerror.Maskf(invalidConfigError, "%T.K8sExtClient must not be empty", config)
-	}
 	if config.Logger == nil {
 		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
 	}
 
-	var crdClient *k8scrdclient.CRDClient
+	// Create a new k8sClient so the legacy scheme can be configured.
```
rossf7: This is a workaround for https://github.com/giantswarm/k8sclient/pull/21#discussion_r352067085 because the controllers use different schemes.

I think we should change to a slice as proposed by Mr @kopiczko. I'd like to work on that tomorrow if there are no objections?

cc @sslavic @tuommaki

tuommaki: No objections. Makes sense.

kopiczko: You can also set it like:

```
AddToScheme: func(s *runtime.Scheme) error {
	err := corev1alpha1.AddToScheme(s)
	if err != nil {
		return microerror.Mask(err)
	}
	err = applicationv1alpha1.AddToScheme(s)
	if err != nil {
		return microerror.Mask(err)
	}

	return nil
},
```

rossf7: Pawel thanks I like that a lot. Done.

service/service.go
```diff
@@ -83,19 +84,30 @@ func New(config Config) (*Service, error) {
...
 	}
 
-	g8sClient, err := versioned.NewForConfig(restConfig)
-	if err != nil {
-		return nil, microerror.Mask(err)
-	}
+	var k8sClient k8sclient.Interface
+	{
+		c := k8sclient.ClientsConfig{
+			AddToScheme: func(s *runtime.Scheme) error {
```
tuommaki: Just noticed something: https://github.com/giantswarm/apiextensions/blob/master/pkg/clientset/versioned/fake/register.go#L37-L59

So this could be:
```suggestion
			AddToScheme: runtime.SchemeBuilder{
				corev1alpha1.AddToScheme,
				applicationv1alpha1.AddToScheme,
			},
```

kopiczko: Sick. How about adding convenience wrapper/methods in k8sclient so we don't have to remember the import?

rossf7: This is  nice and I think maybe the interface can be improved by passing the scheme builder as a parameter.

WIP PR to get feedback is here. https://github.com/giantswarm/k8sclient/pull/23

rossf7:
Tested on `godsmack`. I'll merge in the morning.

tomahawk28:
Take it over from here; merging now.

PR #322, opened 2019-12-04
vendor update; k8sclient

tomahawk28:
Following https://github.com/giantswarm/k8sclient/pull/23
update k8sclient.

PR #325, opened 2019-12-10
Vendor dependencies, update k8s to 1.16.3

tomahawk28:
Following the #322

Will merge after fixing the issue with draughtsman


rossf7:
Tested on gauss. 

Needs to be deployed at the same time as https://github.com/giantswarm/draughtsman/pull/95 as both components share the same tiller.

rossf7:
Tested on geckon. Going with this.

PR #331, opened 2019-12-19
Use priority class created by helmclient in e2e

rossf7:
See https://github.com/giantswarm/helmclient/pull/145

Once this is in helmclient we can drop the 2 helm chart templates. These are only executed for the e2e tests.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -53,7 +53,7 @@ required = [
   name = "github.com/giantswarm/apprclient"
 
 [[constraint]]
-  branch = "master"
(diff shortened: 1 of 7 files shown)


Gopkg.toml
```diff
@@ -53,7 +53,7 @@ required = [
   name = "github.com/giantswarm/apprclient"
 
 [[constraint]]
-  branch = "master"
+  branch = "pull-chart-fixes"
```

rossf7:
Tested on `gorgoth`.

PR #334, opened 2019-12-23
Delete chartconfig controller

rossf7:


rossf7:
Won't work. Controller is needed for the migration process.

PR #336, opened 2019-12-27
Add orphan configmap metric

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/8187

Adds metric for orphan configmaps without a chart CR. Not implemented in app-operator because then it would need to access tenant clusters.

### DIFF
--- a/service/collector/orphan_configmap.go
+++ b/service/collector/orphan_configmap.go
@@ -97,7 +97,7 @@ func (oc *OrphanConfigMap) Collect(ch chan<- prometheus.Metric) error {
 	)
 
 	if len(orphanConfigMaps) > 0 {
-		oc.logger.Log("level", "debug", "message", "found %d orphan configmaps %s", len(orphanConfigMaps), strings.Join(orphanConfigMaps, " "))
+		oc.logger.Log("level", "debug", "message", fmt.Sprintf("found %d orphan configmaps %s", len(orphanConfigMaps), strings.Join(orphanConfigMaps, " ")))
 	}
 
 	oc.logger.Log("level", "debug", "message", "finished collecting metrics for orphan configmaps")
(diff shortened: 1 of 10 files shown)


service/collector/orphan_configmap.go
```diff
@@ -0,0 +1,112 @@
...
+		orphanConfigMapDesc,
+		prometheus.GaugeValue,
+		float64(len(orphanConfigMaps)),
+	)
+
+	if len(orphanConfigMaps) > 0 {
+		oc.logger.Log("level", "debug", "message", fmt.Sprintf("found %d orphan configmaps %s", len(orphanConfigMaps), strings.Join(orphanConfigMaps, " ")))
+	}
+
+	oc.logger.Log("level", "debug", "message", "finished collecting metrics for orphan configmaps")
```
kopiczko: I think that fits better our approach however I'm not sure how we were doing ti so far.

```suggestion
	oc.logger.Log("level", "debug", "message", "collected metrics for orphan configmaps")
```

service/collector/orphan_configmap.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (oc *OrphanConfigMap) Collect(ch chan<- prometheus.Metric) error {
+	oc.logger.Log("level", "debug", "message", "collecting metrics for orphan configmaps")
+
+	charts, err := oc.g8sClient.ApplicationV1alpha1().Charts("").List(metav1.ListOptions{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	desiredConfigMaps := make(map[string]bool)
```
kopiczko: I'd prefer to make it 2 string array instead of constructing custom key. It's also more correct in terms of (unlikely) collisions. See https://play.golang.org/p/QMyCuNv0EO0.

```suggestion
	desiredConfigMaps := make(map[[2]string]bool)
```

service/collector/orphan_configmap.go
```diff
@@ -0,0 +1,112 @@
...
+
+	for _, cr := range charts.Items {
```
kopiczko: `cr` is rather unfortunate name here IMO.

rossf7: Yes this was a bit rough. Fixed.

service/collector/orphan_configmap.go
rossf7: Much better. <3  I think this was outdated I fixed other collectors as well.

service/collector/orphan_configmap.go
rossf7: Thanks that is neat and more correct. Done.

rossf7:
Tested on `ghost`.

PR #337, opened 2019-12-27
Remove unused watch namespace

rossf7:
Some cleanup to remove flag that is no longer used.


rossf7:
Closing in favor of #338

PR #340, opened 2020-01-03
Use updated test charts with K8s 1.16 support

rossf7:


rossf7:
Closing in favor of #342.

PR #342, opened 2020-01-07
Use updated test charts with K8s 1.16 support

rossf7:
Needs this and https://github.com/giantswarm/cnr-server/pull/5


rossf7:
Build is green again. Minor changes. Going with a ping.

PR #344, opened 2020-01-14
Add orphan secrets metric

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/8187

Due to a bug in app-operator we were leaving orphan secrets as well as configmaps. This makes sure both configmaps and secrets are being cleaned up correctly.

### DIFF
--- a/service/collector/orphan_configmap.go
+++ b/service/collector/orphan_configmap.go
@@ -57,8 +57,6 @@ func NewOrphanConfigMap(config OrphanConfigMapConfig) (*OrphanConfigMap, error) 
 }
 
 func (oc *OrphanConfigMap) Collect(ch chan<- prometheus.Metric) error {
-	oc.logger.Log("level", "debug", "message", "collecting metrics for orphan configmaps")
-
 	charts, err := oc.g8sClient.ApplicationV1alpha1().Charts("").List(metav1.ListOptions{})
 	if err != nil {
 		return microerror.Mask(err)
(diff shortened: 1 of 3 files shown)


service/collector/orphan_configmap.go
```diff
@@ -57,8 +57,6 @@ func NewOrphanConfigMap(config OrphanConfigMapConfig) (*OrphanConfigMap, error)
 }
 
 func (oc *OrphanConfigMap) Collect(ch chan<- prometheus.Metric) error {
-	oc.logger.Log("level", "debug", "message", "collecting metrics for orphan configmaps")
-
```
rossf7: Making the logs less noisy. I'll do a follow up PR for the other metrics.

rossf7:
Tested on `ghost`.

service/collector/orphan_secret.go
```diff
@@ -0,0 +1,104 @@
...
+func (oc *OrphanSecret) Collect(ch chan<- prometheus.Metric) error {
+	charts, err := oc.g8sClient.ApplicationV1alpha1().Charts("").List(metav1.ListOptions{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	desiredSecrets := make(map[[2]string]bool)
+
+	for _, chart := range charts.Items {
+		desiredSecrets[[2]string{key.SecretNamespace(chart), key.SecretName(chart)}] = true
```
pipo02mix: cool I did not know it was possible in go :)

rossf7:
Thanks @anvddriesch 👍 and I tested it on ghost earlier.

Aliyun job failed due to connectivity. Going with this.

service/collector/orphan_secret.go
rossf7: This was magic from Mr Pawel.

PR #345, opened 2020-01-14
Remove noisy collector logs

rossf7:
See https://github.com/giantswarm/chart-operator/pull/344/files#r366212863.

I can't remember why but we have extra logging in the collector here. This makes the logs noisy so removing it.

### DIFF
--- a/service/collector/chart_config_resource.go
+++ b/service/collector/chart_config_resource.go
@@ -68,8 +68,6 @@ func NewChartConfigResource(config ChartConfigResourceConfig) (*ChartConfigResou
 }
 
 func (c *ChartConfigResource) Collect(ch chan<- prometheus.Metric) error {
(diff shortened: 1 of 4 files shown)


service/collector/chart_config_resource.go
```diff
@@ -68,8 +68,6 @@ func NewChartConfigResource(config ChartConfigResourceConfig) (*ChartConfigResou
 }
 
 func (c *ChartConfigResource) Collect(ch chan<- prometheus.Metric) error {
-	c.logger.Log("level", "debug", "message", "collecting metrics for ChartConfigs")
```
MarcelMue: IIRC we added these because wiring went belly up at some point and then it was hard to debug why metrics were missing. Fine with me to remove them if you don't find them useful though.

rossf7: Ack thanks for the explanation. I think we're OK now and I compared with aws-operator which doesn't have these.

PR #346, opened 2020-01-14
Fix clusterrole to include priority class

rossf7:
Tiller has not been upgraded in `gorilla` because the cluster role in the chart used by draughtsman was wrong.

### DIFF
--- a/helm/chart-operator-chart/templates/rbac.yaml
+++ b/helm/chart-operator-chart/templates/rbac.yaml
@@ -105,9 +105,9 @@ rules:
   verbs:
   - create
 - apiGroups:
-  - networking.k8s.io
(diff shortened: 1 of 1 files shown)


helm/chart-operator-chart/templates/rbac.yaml
```diff
@@ -105,9 +105,9 @@ rules:
   verbs:
   - create
 - apiGroups:
-  - networking.k8s.io
+  - scheduling.k8s.io
   resources:
-  - networkpolicies
+  - priorityclass
```
rossf7: There were 2 entries for net pols but none for the priority class. The chart used for app CRs is OK.

rossf7:
Tested on `ghost`.

PR #347, opened 2020-01-15
 Make priorityclasses plural in cluster roles

rossf7:
We had some problems earlier with tiller being downgraded and then chart-operator couldn't upgrade it.

I managed to get the rbac wrong in both charts. 🤦‍♂


rossf7:
Tested on `ghost`.

PR #354, opened 2020-02-25
Use client-go in basic integration test

rossf7:
Towards https://github.com/giantswarm/roadmap/issues/30

This is prep for adding a new migration integration test. This is so the legacy chartconfig controller can be removed.

But first sorting out the basic integration test. It uses a helm chart in `apiextensions` to create a chart CR. This is overkill and adds a lot of complexity so now we just use `client-go`.


rossf7:
Trying not to just ask sig-operator on all my reviews. This is for the Helm 3 upgrade which will help with Go modules. Hope thats OK.  🙏

PR #355, opened 2020-02-25
Add chartmigration resource for helm 3

rossf7:
Towards giantswarm/roadmap#30

Adds new chartmigration resource with a new migration integration test.

The migration resource removes the finalizer once cluster-operator has added an annotation to signal that the migration is complete.

With the new migration resource the legacy chartconfig controller can be disabled. It will be removed in follow up PRs.


rossf7:
Closing in favor of #358

PR #356, opened 2020-02-26
Add migration integration test

rossf7:
Towards giantswarm/roadmap#30

Adds new migration integration test. Ensures that chartconfig CRs can be migrated to chart CRs without affecting the related helm release.

This is prep for removing the legacy chartconfig controller. Its only purpose is deleting chartconfig CRs as part of the migration.


integration/test/chart/migration/migration_test.go
```diff
@@ -0,0 +1,226 @@
...
+		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("checking release %#q is deployed", key.TestAppReleaseName()))
+
+		err := config.Release.WaitForStatus(ctx, key.TestAppReleaseName(), "DEPLOYED")
+		if err != nil {
+			t.Fatalf("expected %#v got %#v", nil, err)
+		}
+
+		config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("release %#q is deployed", key.TestAppReleaseName()))
+	}
+}
```
tomahawk28: Just to confirm; Does it mean the status of the app would be `FAILED` or something else if `chart-operator` do not have migration resources?

rossf7: Good to check. There should be a helm release called `test-app`. At the end of the test it should still have status `DEPLOYED`. 

If it got deleted or was `FAILED` then the test will fail.

rossf7:
I don't think test is controversial. Going with this.

tomahawk28:
Sorry, I just got busy with other things.

rossf7:
@tomahawk28 No worries. ❤️ 

I got an error trying to add my comment due to the outage so I gave up.

PR #358, opened 2020-02-28
Add chartmigration resource and disable chartconfig controller

rossf7:
Towards giantswarm/roadmap#30

Now that we only use chart CRs in new tenant clusters the only purpose of the legacy chartconfig controller is for deleting chartconfig CRs.

This is now done by a new chartmigration resource. It checks for the `chart-operator.giantswarm.io/delete-custom-resource-only` annotation which is added by cluster-operator once the migration is complete.

https://github.com/giantswarm/cluster-operator/blob/legacy/service/controller/resource/appmigration/create.go#L150-L155


.circleci/config.yml
```diff
@@ -94,30 +94,6 @@ jobs:
...
-      E2E_TEST_DIR: "integration/test/chartconfig/chartvalues"
-      E2E_TESTED_VERSION: "current"
-    <<: *e2eTest
-
-  e2eTestChartConfigCurChartValuesMaster:
-    environment:
-      E2E_TEST_DIR: "integration/test/chartconfig/chartvalues"
-      E2E_TESTED_VERSION: "current"
-    <<: *e2eTest
-
```
rossf7: chartconfig tests get removed.

service/controller/chart/v1/resource/chartmigration/create.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding chartconfig %#q", cr.Name))
+
+	chartConfig, err := r.g8sClient.CoreV1alpha1().ChartConfigs(cr.Namespace).Get(cr.Name, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find chartconfig %#q. nothing to do.", cr.Name))
+		return nil
```
rossf7: If there is no chartconfig CR we can stop here.

service/controller/chart/v1/resource/chartmigration/create.go
```diff
@@ -0,0 +1,60 @@
...
+	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("checking if chartconfig %#q has been migrated", cr.Name))
+
+	if key.HasDeleteCROnlyAnnotation(chartConfig) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("chartconfig %#q has been migrated", cr.Name))
+		r.logger.LogCtx(ctx, "level", "debug", "message", "removing finalizer")
+
+		finalizers := []string{}
+
+		for _, f := range chartConfig.ObjectMeta.Finalizers {
+			if f == "operatorkit.giantswarm.io/chart-operator-chartconfig" {
```
rossf7: Once the annotation is in place we only remove the operatorkit finalizer.

service/controller/chart/v1/resource/chartmigration/create.go
```diff
@@ -0,0 +1,60 @@
...
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("removing finalizer %#q", f))
+			} else {
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("keeping finalizer %#q", f))
+				finalizers = append(finalizers, f)
+			}
+		}
+
+		chartConfig.ObjectMeta.Finalizers = finalizers
+		_, err = r.g8sClient.CoreV1alpha1().ChartConfigs("giantswarm").Update(chartConfig)
```
tomahawk28: Can we simply use patch operation to remove specific finalizers? I think we did many times before.

rossf7: True good point. I'll update to do that. 

I also found a problem with getting the chartconfig CR. Some of the CR names are different as they have `kubernetes-` prefixes. 

But they have the same app label so I'll use that instead.

rossf7:
Changes made. Tested on `ghost`.

rossf7:
@tomahawk28 PTAL. Doing a final test but this should be OK now.

rossf7:
Tested on `ghost`. Going with this.

PR #359, opened 2020-02-28
Adding relevant reason for an initial failure

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/9151

Improving for debugging initial failure cases. With this change, you don't need to dig into logs but only checking CR's status would fine.


tomahawk28:
After: 
```
status:
  appVersion: ""
  reason: 'helm error: (rpc error: code = Unknown desc = validation failed: [unable
    to recognize "": no matches for kind "PodSecurityPolicy" in version "extensions/v1beta1",
    unable to recognize "": no matches for kind "Deployment" in version "extensions/v1beta1"])'
  release:
    lastDeployed: "0001-01-01T00:00:00Z"
    status: Not installed
  version: ""
```

tomahawk28:
Tested on `gauss`

tomahawk28:
Sure, I will look into it now.

PR #362, opened 2020-03-05
Do not block when creating or updating releases

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/9246

This is to prevent blocking from long running or failing charts. I've tested on ginger in both control plane and tenants and looks OK.

### DIFF
--- a/service/controller/chart/v1/resource/release/create.go
+++ b/service/controller/chart/v1/resource/release/create.go
@@ -3,6 +3,7 @@ package release
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/giantswarm/helmclient"
 	"github.com/giantswarm/microerror"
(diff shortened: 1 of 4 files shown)


service/controller/chart/v1/resource/release/create.go
```diff
@@ -70,9 +71,30 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
...
+			// We need to pass the ValueOverrides option to make the install process
+			// use the default values and prevent errors on nested values.
+			err = r.helmClient.InstallReleaseFromTarball(ctx, tarballPath, ns, helm.ReleaseName(releaseState.Name), helm.ValueOverrides(releaseState.ValuesYAML))
+			close(ch)
+		}()
+
+		select {
+		case <-ch:
+			// Fall through.
+		case <-time.After(3 * time.Second):
```
rossf7: We use time.After and not context.WithTimeout because it doesn't work with our k8sportforward lib.

The timeout is short but many of our simple charts install quickly and then there is no change in behavior.

rossf7:
There is a problem processing the gatekeeper rules chart. The CR status is staying in `PENDING_INSTALL`. Added some debug.

rossf7:
Tested on `ghost`.

Did more testing. gatekeeper rules chart installs OK it just takes some time. Going with this.

PR #365, opened 2020-03-09
Delete chartconfig controller and related code

rossf7:
Towards giantswarm/roadmap#30

The legacy chartconfig controller was disabled in https://github.com/giantswarm/chart-operator/pull/358. This now deletes the code.


rossf7:
Tested on `ghost`.

PR #366, opened 2020-03-09
Add missing version bundles

rossf7:


service/version_bundles.go
```diff
@@ -4,12 +4,20 @@ import (
...
+	chartconfigv6 "github.com/giantswarm/chart-operator/service/controller/chartconfig/v6"
+	chartconfigv7 "github.com/giantswarm/chart-operator/service/controller/chartconfig/v7"
 )
 
 func NewVersionBundles() []versionbundle.Bundle {
 	var versionBundles []versionbundle.Bundle
 
 	versionBundles = append(versionBundles, chartv1.VersionBundle())
 
+	versionBundles = append(versionBundles, chartconfigv5.VersionBundle())
```
tomahawk28: Don't we deprecate this `chartconfig` from the latest version?

rossf7: chart-operator is used in the releases with 0.7.0 version. e.g https://github.com/giantswarm/releases/blob/c0dfb19574adf79217e65b6c09df0f33dee1d377/aws.yaml#L115

Without the endpoint these releases don't show up. Getting that fixed and then will see what the long term fix is.

PR #351, opened 2020-02-05
migrate from dep to Go modules

kopiczko:

PR #369, opened 2020-03-12
helm3: Migrate from dep to go modules

rossf7:
Towards https://github.com/giantswarm/roadmap/issues/30

We now have a `helm3` branch that we will use for the rest of the migration. This is based on https://github.com/giantswarm/chart-operator/pull/351 from Pawel.

Once the branch is using modules I'll introduce the helm3 branch of helmclient.


go.mod
```diff
@@ -0,0 +1,77 @@
...
+	k8s.io/kube-controller-manager v0.0.0 => k8s.io/kube-controller-manager v0.16.6
+	k8s.io/kube-proxy v0.0.0 => k8s.io/kube-proxy v0.16.6
+	k8s.io/kube-scheduler v0.0.0 => k8s.io/kube-scheduler v0.16.6
+	k8s.io/kubectl v0.0.0 => k8s.io/kubectl v0.16.6
+	k8s.io/kubelet v0.0.0 => k8s.io/kubelet v0.16.6
+	k8s.io/legacy-cloud-providers v0.0.0 => k8s.io/legacy-cloud-providers v0.16.6
+	k8s.io/metrics v0.0.0 => k8s.io/metrics v0.16.6
+	k8s.io/sample-apiserver v0.0.0 => k8s.io/sample-apiserver v0.16.6
+	k8s.io/utils v0.0.0 => k8s.io/utils v0.0.0-20191114200735-6ca3b61696b6
+)
```
kopiczko: Do we need this replace BS when we don't have dependency on k8s.io/kubernetes?

rossf7: I can check but this is just an intermediate step before adding the helm3 branch.

The dependencies then get radically simpler. See https://github.com/giantswarm/helmclient/pull/178.

kopiczko: Ack. Ignore me then.

rossf7: Yes situation will improve soon. Going with this.

PR #372, opened 2020-03-12
helm3:  Drop unit tests except desired state

rossf7:
Towards giantswarm/roadmap#30

Dropping the unit tests except for desired state. They are not useful and just get in the way.

We have pretty good integration test coverage and we should keep on improving that.


rossf7:
I don't think this is controversial. Going with a ping.

PR #373, opened 2020-03-12
helm3:  Upgrade helmclient to Helm 3

rossf7:
Towards giantswarm/roadmap#30

PR #375, opened 2020-03-16
Update chart for Helm 3 and linter

rossf7:
Towards giantswarm/roadmap#30

Breaking down the big ball of mud in https://github.com/giantswarm/chart-operator/pull/373.

### DIFF
--- a/helm/chart-operator/templates/deployment.yaml
+++ b/helm/chart-operator/templates/deployment.yaml
@@ -20,7 +20,7 @@ spec:
         app: {{ .Values.project.name }}
         giantswarm.io/service-type: "managed"
       annotations:
-        releasetime: {{ $.Release.Time }}
+        releaseRevision: {{ $.Release.Revision | quote }}
     spec:
       affinity:
         nodeAffinity:
--- a/helm/chart-operator/templates/priorityclass.yaml
+++ b/helm/chart-operator/templates/priorityclass.yaml
@@ -1,0 +1,9 @@
+{{- if .Values.e2e }}
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: giantswarm-critical
+value: 1000000000
+globalDefault: false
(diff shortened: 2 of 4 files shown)


helm/chart-operator/templates/deployment.yaml
```diff
@@ -20,7 +20,7 @@ spec:
         app: {{ .Values.project.name }}
         giantswarm.io/service-type: "managed"
       annotations:
-        releasetime: {{ $.Release.Time }}
+        releaseRevision: {{ $.Release.Revision | quote }}
```
rossf7: Release.Time got removed in Helm 3 but Release.Revision still exists.
https://github.com/helm/helm/issues/5732

We use this so pods are rolled when the release is updated. So revision also works.

helm/chart-operator/templates/priorityclass.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if .Values.e2e }}
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: giantswarm-critical
+value: 1000000000
+globalDefault: false
+description: "This priority class is used by giantswarm kubernetes components."
+{{- end }}
```
rossf7: Since helmclient will no longer create the priority class we need this template for the tests.

helm/chart-operator/templates/priorityclass.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if .Values.e2e }}
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
```
tomahawk28: Even though it's e2e-only resource, I would like to put labels as we prefer. https://github.com/giantswarm/fmt/blob/master/kubernetes/annotations_and_labels.md#common-labels

rossf7: Good point. Done.

PR #376, opened 2020-03-16
helm3: Remove usage of e2e-harness in integration tests

rossf7:
Towards giantswarm/roadmap#30

e2e-harness uses helmclient. So this kills 2 birds with one stone by moving the setup code into the operator.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,9 +5,8 @@ go 1.13
 require (
 	github.com/chai2010/gettext-go v0.0.0-20191225085308-6b9f4b1008e1 // indirect
 	github.com/giantswarm/apiextensions v0.1.0
-	github.com/giantswarm/apprclient v0.0.0-20191209123802-955b7e89e6e2 // indirect
+	github.com/giantswarm/appcatalog v0.1.11
 	github.com/giantswarm/backoff v0.0.0-20190913091243-4dd491125192
-	github.com/giantswarm/e2e-harness v0.1.1-0.20191209145255-b2048d8645c1
 	github.com/giantswarm/e2esetup v0.0.0-20191209131007-01b9f9061692
 	github.com/giantswarm/exporterkit v0.0.0-20190619131829-9749deade60f
 	github.com/giantswarm/helmclient v0.0.0-20200205160527-61453b0b25de
--- a/integration/release/error.go
+++ b/integration/release/error.go
@@ -1,0 +1,32 @@
+package release
+
+import (
+	"github.com/giantswarm/microerror"
+)
+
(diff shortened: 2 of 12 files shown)


go.mod
```diff
@@ -5,9 +5,8 @@ go 1.13
 require (
 	github.com/chai2010/gettext-go v0.0.0-20191225085308-6b9f4b1008e1 // indirect
 	github.com/giantswarm/apiextensions v0.1.0
-	github.com/giantswarm/apprclient v0.0.0-20191209123802-955b7e89e6e2 // indirect
+	github.com/giantswarm/appcatalog v0.1.11
 	github.com/giantswarm/backoff v0.0.0-20190913091243-4dd491125192
-	github.com/giantswarm/e2e-harness v0.1.1-0.20191209145255-b2048d8645c1
```
rossf7: 🔥

integration/release/error.go
```diff
@@ -0,0 +1,32 @@
+package release
```
rossf7: We only need `WaitForChartStatus` and `WaitForChartInfo`. So moving this in from e2e-harness.

integration/setup/config.go
```diff
@@ -15,9 +16,9 @@ const (
 )
 
 type Config struct {
-	HelmClient *helmclient.Client
+	HelmClient helmclient.Interface
```
rossf7: Ninja-ing this in.

integration/setup/setup.go
```diff
@@ -40,24 +45,85 @@ func installResources(ctx context.Context, config Config) error {
...
 
 	{
-		err := config.K8s.EnsureNamespaceCreated(ctx, namespace)
+		err := config.K8s.EnsureNamespaceCreated(ctx, key.Namespace())
 		if err != nil {
 			return microerror.Mask(err)
 		}
 	}
 
+	var latestOperatorRelease string
```
rossf7: Instead of using `Release.InstallOperator` we just install it here. This will get simpler once the operator is flat as we can use `project.Version()`.

integration/test/chart/basic/basic_test.go
```diff
@@ -35,19 +35,19 @@ func TestChartLifecycle(t *testing.T) {
 		cr := &v1alpha1.Chart{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      key.TestAppReleaseName(),
-				Namespace: "giantswarm",
+				Namespace: key.Namespace(),
```
rossf7: All Helm 3 actions are scoped to a namespace. So this is prep for that.

PR #379, opened 2020-03-17
Helm3 migration job for chart-operator

tomahawk28:
helm3 job migration for chart-operator. 

Spec: https://github.com/giantswarm/giantswarm/blob/master/areas/managed-services/specs/helm-3-upgrade.md


helm/chart-operator/values.yaml
```diff
@@ -7,8 +7,11 @@ e2e: false
 
 externalDNSIP: 8.8.8.8
 
+global:
+  image:
+    registry: quay.io
```
tomahawk28: This is for having same registry as a parent chart.

helm/chart-operator/templates/deployment.yaml
```diff
@@ -20,7 +20,7 @@ spec:
         app: {{ .Values.project.name }}
         giantswarm.io/service-type: "managed"
       annotations:
-        releaseRevision: {{ $.Release.Revision | quote }}
+        releasetime: {{ $.Release.Time }}
```
tomahawk28: Same fix as Ross did

PR #383, opened 2020-03-20
dep: Use released versions of libraries

rossf7:
Using the released versions of the apps. To get us up to date.

This also uncovered some problems with the OpenAPI validation. Fixed in https://github.com/giantswarm/apiextensions/pull/364.


rossf7:
Tested on `geckon`. Vendor PR going with a ping.

PR #384, opened 2020-03-23
Revert "dep: Use released versions of libraries (#383)"

rossf7:
This reverts commit c0b527fd7cfdc9ec9670a906b3ec6870ac563a68.

This doesn't work in anubis because its still on k8s 1.14.


rossf7:
Going with a ping.

PR #385, opened 2020-03-23
Install helm 2to3 migration

tomahawk28:
helm3 job migration for chart-operator.

installing https://github.com/giantswarm/helm-2to3-migration on releasemigration, only if there is dangling helm v2 releases.

Spec: https://github.com/giantswarm/giantswarm/blob/master/areas/managed-services/specs/helm-3-upgrade.md


tomahawk28:
Tested on `gauss`

```
 ~  helm3 --namespace kube-system ls
NAME              	NAMESPACE  	REVISION	UPDATED                                	STATUS  	CHART                       	APP VERSION
cert-exporter     	kube-system	1       	2020-03-23 13:57:00.39320944 +0000 UTC 	deployed	cert-exporter-1.2.1
cert-manager      	kube-system	2       	2020-03-23 17:01:18.647126436 +0000 UTC	deployed	cert-manager-app-1.0.6      	0.9.0
cluster-autoscaler	kube-system	1       	2020-03-23 14:01:05.682215139 +0000 UTC	deployed	cluster-autoscaler-app-1.1.4
coredns           	kube-system	1       	2020-03-23 14:01:21.05564395 +0000 UTC 	deployed	coredns-app-1.1.8           	1.6.5
external-dns      	kube-system	1       	2020-03-23 13:56:59.793118557 +0000 UTC	deployed	external-dns-app-1.2.0      	v0.5.18
kiam              	kube-system	2       	2020-03-23 17:01:18.205982844 +0000 UTC	deployed	kiam-app-1.2.1              	3.5
kube-state-metrics	kube-system	1       	2020-03-23 14:01:16.175983707 +0000 UTC	deployed	kube-state-metrics-app-1.0.4	v1.9.2
metrics-server    	kube-system	1       	2020-03-23 14:01:09.276536556 +0000 UTC	deployed	metrics-server-app-1.0.0    	0.3.3
net-exporter      	kube-system	1       	2020-03-23 13:57:01.993272038 +0000 UTC	deployed	net-exporter-1.7.0
node-exporter     	kube-system	1       	2020-03-23 14:01:23.655379615 +0000 UTC	deployed	node-exporter-app-1.2.0     	v0.18.1
 ~  helm3 --namespace giantswarm ls
NAME               	NAMESPACE 	REVISION	UPDATED                                	STATUS  	CHART                                                         	APP VERSION
chart-operator     	giantswarm	2       	2020-03-23 17:01:19.620315384 +0000 UTC	deployed	chart-operator-0.12.1-c3f005f01c473703f66bece2a84506e4cb0aed2b	0.12.1-c3f005f01c473703f66bece2a84506e4cb0aed2b
helm-2to3-migration	giantswarm	1       	2020-03-23 17:01:12.898262505 +0000 UTC	deployed	helm-2to3-migration-1.0.1                                     	0.5.0
```

service/controller/chart/v1/resource/releasemigration/resource.go
```diff
@@ -53,3 +70,103 @@ func New(config Config) (*Resource, error) {
...
+	releases := make([]string, 0, len(hadReleases))
+	for k := range hadReleases {
+		releases = append(releases, k)
+	}
+
+	return releases, nil
+}
+
+func (r *Resource) ensureReleasesMigrated(ctx context.Context) error {
+	// 0. Found all dangling helm release v2
```
tomahawk28: Do these comments help? I think so; would like to preserve it.

service/controller/chart/v1/resource/releasemigration/resource.go
```diff
@@ -53,3 +70,103 @@ func New(config Config) (*Resource, error) {
...
+	cms, err := r.k8sClient.CoreV1().ConfigMaps(r.tillerNamespace).List(lo)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	hadReleases := map[string]bool{}
+	for _, cm := range cms.Items {
+		name, _ := cm.GetLabels()["NAME"]
+		if _, ok := hadReleases[name]; !ok {
+			hadReleases[name] = true
```
tomahawk28: Only gathering unique release name by using map.

service/controller/chart/v1/resource/releasemigration/resource.go
```diff
@@ -53,3 +70,103 @@ func New(config Config) (*Resource, error) {
...
+	releases, err := r.findHelmV2Releases(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// 1. Install helm-2to3-migration app
+	{
+		var tarballPath string
+		{
+			tarballURL, err := appcatalog.GetLatestChart(ctx, key.DefaultCatalogStorageURL(), "helm-2to3-migration")
```
tomahawk28: Getting the latest app by using `appcatalog` 🎉

service/controller/chart/v1/resource/releasemigration/resource.go
rossf7:
```suggestion
	// Find all helm v2 releases.
```

service/controller/chart/v1/resource/releasemigration/resource.go
```diff
@@ -53,3 +70,103 @@ func New(config Config) (*Resource, error) {
...
+}
+
+func (r *Resource) ensureReleasesMigrated(ctx context.Context) error {
+	// 0. Found all dangling helm release v2
+	releases, err := r.findHelmV2Releases(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// 1. Install helm-2to3-migration app
```
rossf7:
```suggestion
	// Install helm-2to3-migration app.
```

service/controller/chart/v1/resource/releasemigration/create.go
```diff
@@ -41,8 +41,15 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	// It means helm v3 release migration is not started.
 	if hasConfigMap && !hasSecret {
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("release %#q helmV3 migration not started", key.ReleaseName(cr)))
-		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation")
-		reconciliationcanceledcontext.SetCanceled(ctx)
+		r.logger.LogCtx(ctx, "level", "debug", "message", "installing helm-2to3-migration")
```
rossf7:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("installing %#q", migrationApp)
```

We need the release name in a few places. So should we have a constant?

service/controller/chart/v1/resource/releasemigration/resource.go
rossf7: Agree but the numbering is not our usual comments style.

service/controller/chart/v1/resource/releasemigration/resource.go
```diff
@@ -53,3 +70,103 @@ func New(config Config) (*Resource, error) {
...
+			}
+
+			err = r.helmClient.InstallReleaseFromTarball(ctx, tarballPath, r.tillerNamespace, values, opts)
+			if err != nil {
+				return microerror.Mask(err)
+			}
+		}
+	}
+
+	// 2. Wait until all helm v2 release are deleted
```
rossf7:
```suggestion
	// Wait until all helm v2 releases are deleted.
```

service/controller/chart/v1/resource/releasemigration/resource.go
```diff
@@ -53,3 +70,103 @@ func New(config Config) (*Resource, error) {
...
+	operation := func() error {
```
rossf7:
```suggestion
	o := func() error {
```

What about using the shorter var name here?

service/controller/chart/v1/resource/releasemigration/resource.go
```diff
@@ -53,3 +70,103 @@ func New(config Config) (*Resource, error) {
...
+			return microerror.Mask(err)
+		}
+		if len(releases) > 0 {
+			return microerror.Maskf(releasesNotDeletedError, "helm v2 releases not deleted: %#q", releases)
+		}
+
+		return nil
+	}
+
+	notify := func(err error, t time.Duration) {
```
rossf7:
```suggestion
	n := func(err error, t time.Duration) {
```

Same here.

service/controller/chart/v1/resource/releasemigration/resource.go
```diff
@@ -53,3 +70,103 @@ func New(config Config) (*Resource, error) {
...
+		r.logger.Log("level", "debug", "message", "failed to deleted all helm v2 releases")
+	}
+
+	b := backoff.NewConstant(backoff.ShortMaxWait, backoff.ShortMaxInterval)
```
rossf7: I like using the actual time intervals in backoffs. I always forget the constant values so for me its more readable.

The migration is an unusual situation but I have a worry about using a backoff for this. Could we cancel the reconciliation and check again in the next loop?

tomahawk28: Mostly in the next loop, it will reconcile on the next chart CRs, and would try to gather helm v2 release and would attempt to install `helm-2to3-migration` app. 
I would need to put an error handling guessing on this situation, but does it worth than this backoff? I would like to keep on this backoff for this reason. 

And What do you think having 3 seconds for interval and 30 seconds for the maximum wait?

rossf7: OK cool I'm fine with keeping the backoff. 👍 I think those settings are good. 

But the migration may go wrong. Maybe we should just check if the migration app release exists?

If so we can cancel the resource. We will get alerted because the migration did not complete. WDYT?

tomahawk28: > We will get alerted because the migration did not complete. WDYT? 

If we canceled the resource, we get paged? Don't we need to put an unhandled error?

rossf7: No we need to add a new metric for un-migrated releases. We can alert based on that.

https://github.com/giantswarm/giantswarm/blob/master/areas/managed-services/specs/helm-3-upgrade.md#non-migrated-helm-releases-metric

service/controller/chart/v1/resource/releasemigration/create.go
```diff
@@ -45,7 +45,12 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 
 		// install helm-2to3-migration app
 		err := r.ensureReleasesMigrated(ctx)
-		if err != nil {
+		if IsReleaseAlreadyExists(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("release %#q already exists", migrationApp))
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation")
+			reconciliationcanceledcontext.SetCanceled(ctx)
```
tomahawk28: I think canceling reconciliation makes sense since we don't know that migration was done at this point.

service/controller/chart/v1/resource/releasemigration/resource.go
tomahawk28: Thanks, May I can start from the next PR?

rossf7: Yes definitely. Better in a separate PR.

PR #386, opened 2020-03-24
Add helm v2 release collector

tomahawk28:
Adding additional metrics for remaining helm v2 releases.

Spec: https://github.com/giantswarm/giantswarm/blob/master/areas/managed-services/specs/helm-3-upgrade.md#non-migrated-helm-releases-metric

### DIFF
--- a/service/collector/helm_v2_release.go
+++ b/service/collector/helm_v2_release.go
@@ -1,0 +1,100 @@
+package collector
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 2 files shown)


service/collector/helm_v2_release.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+var (
+	helmV2ReleaseDesc *prometheus.Desc = prometheus.NewDesc(
+		prometheus.BuildFQName(Namespace, "helm", "release"),
+		"Dangling Helm V2 releases",
+		[]string{},
```
tomahawk28: I can't think of any labels belongs to here.

PR #393, opened 2020-03-31
Rebase master helm3

tomahawk28:


go.mod
```diff
@@ -3,25 +3,54 @@ module github.com/giantswarm/chart-operator
...
+	k8s.io/apimachinery v0.17.2
+	k8s.io/client-go v0.17.2
 	sigs.k8s.io/yaml v1.1.0
 )
+
+// Workaround for https://github.com/sirupsen/logrus/issues/570
+// Additional reading: https://github.com/golang/go/issues/28489
+// Solution: https://github.com/moby/moby/issues/39302
+replace github.com/docker/docker => github.com/docker/engine v1.4.2-0.20191113042239-ea84732a7725
+
```
rossf7: We shouldn't need all this replace stuff.

Its only needed for Helm 2 because it imports `k8s.io/kubernetes`. This branch is already using helm 3.

PR #394, opened 2020-03-31
Add initial dependabot config

stone-z:


.dependabot/config.yml
```diff
@@ -0,0 +1,9 @@
+version: 1
+update_configs:
+  - package_manager: "go:modules"
+    directory: "/"
+    update_schedule: "daily"
+    ignored_updates:
+      - match:
+          # Ignore k8s.io dependencies
+          dependency_name: "k8s.io/*"
```
tomahawk28: perhaps can you add a newline?

stone-z: :+1:

.dependabot/config.yml
```diff
@@ -0,0 +1,9 @@
+version: 1
+update_configs:
+  - package_manager: "go:modules"
+    directory: "/"
+    update_schedule: "daily"
+    ignored_updates:
```
tomahawk28: Sorry, I think it's better to have `allowed_updates` with `github.com/giantswarm/*`. but we can change after this.

stone-z: IMO external updates are as important if not more important than internal ones in general. Of course you can tune this, but external components introduce vulnerabilities and these should be monitored

rossf7: Yes happy to try it with just `k8s.io` ignored. We can iterate if we need to add more later.

PR #396, opened 2020-04-03
Flatten chart controller

rossf7:


helm/chart-operator/values.yaml
```diff
@@ -1,8 +1,5 @@
 clusterDNSIP: 172.31.0.10
 
-cnr:
-  address: https://quay.io
-
```
rossf7, line 5: No longer needed.

pkg/project/project.go
```diff
@@ -2,12 +2,19 @@ package project
 
 var (
 	description = "chart-operator is our agent for deploying Helm charts as releases."
-	name        = "chart-operator"
 	gitSHA      = "n/a"
+	name        = "chart-operator"
 	source      = "https://github.com/giantswarm/chart-operator"
-	version     = "n/a"
+	version     = "1.0.0-alpha.1-dev"
```
rossf7, line 8: This is a bit of a monster but we created a alpha1 release to make testing easier.

pkg/project/project.go
```diff
@@ -2,12 +2,19 @@ package project
...
 )
 
+// ChartVersion is fixed for chart CRs. This is because they exist in both
+// control plane and tenant clusters and their version is not linked to a
+// release. We may revisit this in future.
+func ChartVersion() string {
+	return "1.0.0"
+}
```
rossf7, line 16: This is the controversial part but I don't see an alternative currently. As we have multiple components generating app CRs.

tomahawk28: If we roll out 1.0.0, how can we avoid `MultipleOperatorsRunningSameBundleVersion` alerts from CPs?

rossf7: When we add it to the app collections we need to set the unique param to true. So we only run one chart-operator per control plane.

e.g. https://github.com/giantswarm/gatekeeper-app/blob/a47e6a30f1f570af9e66bfa7cc859184308fcf45/.circleci/config.yml#L48

kopiczko: I think it's oke when we run it as an unique app. Maybe `0.0.0` would be better indication that we don't really version it?

rossf7: I think `0.0.0` is maybe clearer but we will have the same problem in app-operator and the app CRs are user facing and in docs.

How can we move this forward? As we need to continue with the flattening.

kopiczko: I see the UX problem there, but I don't see a problem with merging this PR. You can go with 1.0.0 for now and maybe ask appropriate people if 0.0.0 in CRs label values is not confusing. I'd prefer later one because it isn't that confusing for devs.

rossf7: Thanks makes sense. Going with this but will see if we can find another solution for the labels.

PR #399, opened 2020-04-06
helm3: Remove appr chart and update Circle config

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7896

Towards the flattening to remove the chart used by draughtsman and related logic in the circle config.


.circleci/config.yml
```diff
@@ -16,41 +16,10 @@ jobs:
     - store_test_results:
         path: /tmp/results
     - run: ./architect publish --pipeline=false --channels=${CIRCLE_SHA1}
-    - run: |
-        sed -i 's/version:.*/version: 1.0.0-'"${CIRCLE_SHA1}"'/' helm/chart-operator-chart/Chart.yaml
-        ./architect publish --pipeline=false --channels=${CIRCLE_SHA1}
```
rossf7, line 21: Removing all the logic for deploying with appr / draughtsman.

.circleci/config.yml
```diff
@@ -86,13 +55,16 @@ workflows:
...
-          image: "registry-intl.cn-shanghai.aliyuncs.com/giantswarm/chart-operator"
-          username_envar: "ALIYUN_USERNAME"
-          password_envar: "ALIYUN_PASSWORD"
+      - architect/push-to-app-catalog:
+          name: push-chart-operator-to-control-plane-app-catalog
+          app_catalog: "control-plane-catalog"
+          app_catalog_test: "control-plane-test-catalog"
+          chart: "chart-operator"
           requires:
-            - build
```
rossf7, line 95: No need for legacy docker image.

.circleci/config.yml
```diff
@@ -86,13 +55,16 @@ workflows:
...
+            - push-chart-operator-to-quay
```
rossf7, line 64: We should install chart-operator in control planes from its catalog. So now its pushed to both control-plane-catalog and default-catalog.

server/server.go
```diff
@@ -80,7 +80,6 @@ type Server struct {
 	// Internals
 	bootOnce     sync.Once
 	config       microserver.Config
-	serviceName  string
```
rossf7, line 83: The rest of this is dropping dead code and fixing go vet issues.

.circleci/config.yml
rossf7: I'd like to add the app collection wiring later once we're ready to merge into master. As if the Helm 3 version of chart-operator gets deployed too early the migration logic will cause a mess.

tomahawk28:
Found https://github.com/giantswarm/default-test-catalog/blob/master/chart-operator-1.0.0-alpha.1-3d9ed2e29a364badb4780a75621b5aa86d835771.tgz 

LGTM :)

PR #408, opened 2020-04-09
helm3: Make HTTP client timeout in helmclient configurable

rossf7:
Porting https://github.com/giantswarm/chart-operator/pull/402 to helm3 branch to keep it in sync.


rossf7:
Getting helm3 branch in sync. Going with a ping.

PR #409, opened 2020-04-09
helm3: Fix update state calculation and status resource for long running deploys

rossf7:
Porting https://github.com/giantswarm/chart-operator/pull/406 to helm3 branch.


rossf7:
Getting helm3 branch in sync. Going with a ping.

PR #410, opened 2020-04-09
Update orb version and set AppVersion

rossf7:
Tagging 0.12.2 earlier failed I think due to the AppVersion templating.

Also updating architect-orb to latest so we use medium Circle VMs in CI to reduce annoying KIND cluster failures.


rossf7:
Going with a ping and will retry releasing 0.12.1.

PR #411, opened 2020-04-09
Release 0.12.3

rossf7:
Creating a new release rather than trying to retag 0.12.2 which will likely fail.


rossf7:
Another ping.

PR #415, opened 2020-04-13
helm3: Delete releasemigration resource

rossf7:
Got added in error when code was flattened. Resource has moved to app-operator.


rossf7:
Cleaning up code reintroduced in error. Going with a ping.

PR #418, opened 2020-04-14
Remove v1 resourceset

rossf7:

Completes go part of the flattening. Next will be helm chart changes.


service/controller/chart/chart.go
```diff
@@ -11,10 +11,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/giantswarm/chart-operator/pkg/project"
-	v1 "github.com/giantswarm/chart-operator/service/controller/chart/v1"
 )
 
-const chartConfigControllerSuffix = "-chart"
```
rossf7, line 17: Fixing the name.

pkg/project/project.go
```diff
@@ -5,9 +5,16 @@ var (
 	name        = "chart-operator"
 	gitSHA      = "n/a"
 	source      = "https://github.com/giantswarm/chart-operator"
-	version     = "n/a"
```
rossf7, line 8: Current version is 0.12.3 but going to bump it to 1.0. As per the flattening docs and Helm 3 is coming soon.

pkg/project/project.go
```diff
@@ -5,9 +5,16 @@ var (
...
+	version     = "1.0.0"
 )
 
+// ChartVersion is fixed for chart CRs. This is because they exist in both
+// control plane and tenant clusters and their version is not linked to a
+// release. We may revisit this in future.
+func ChartVersion() string {
+	return "1.0.0"
+}
```
rossf7, line 16: Same approach we're using in the helm3 branch.

rossf7:
Tested on `geckon`.

PR #419, opened 2020-04-14
Set chart metadata using helper

rossf7:

Sets metadata as per https://github.com/giantswarm/fmt/blob/f657e3286f12f86d0bb9f439e065cf1d10ff9fc6/kubernetes/annotations_and_labels.md#labels-set-on-objects-installed-by-charts.


helm/chart-operator/templates/deployment.yaml
```diff
@@ -4,23 +4,21 @@ metadata:
...
   strategy:
     type: RollingUpdate
   template:
     metadata:
-      labels:
-        app: {{ .Values.project.name }}
-        giantswarm.io/service-type: "managed"
       annotations:
-        releasetime: {{ $.Release.Time }}
+        releaseRevision: {{ $.Release.Revision | quote }}
```
rossf7, line 19: `Release.Time` is removed in Helm 3. So this preps the chart.

helm/chart-operator/templates/priorityclass.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: giantswarm-critical
+  labels:
+    {{- include "chart-operator.labels" . | nindent 4 }}
+value: 1000000000
+globalDefault: false
+description: "This priority class is used by giantswarm kubernetes components."
+{{- end }}
```
rossf7, line 11: More prep for helm 3. As we don't install tiller we need to create the priority class in the chart for integration tests.

helm/chart-operator/templates/_helpers.tpl
```diff
@@ -0,0 +1,35 @@
...
+{{- define "chart-operator.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Common labels
+*/}}
+{{- define "chart-operator.labels" -}}
+app: {{ include "chart-operator.name" . | quote }}
+{{ include "chart-operator.selectorLabels" . }}
```
kopiczko, line 21: Let's move this line to top or bottom of the section. I think it make more sense to care about alphabetical order during development that after rendering (as the later is impossible with inclusions).

helm/chart-operator/templates/priorityclass.yaml
kopiczko: For my interest: why we don't need this with tiller?

helm/chart-operator/templates/_helpers.tpl
rossf7: Thanks good point. I moved to the top since the selector labels are important.

helm/chart-operator/templates/priorityclass.yaml
rossf7: The priority class is one of several created by EnsureTillerInstalled. So when it gets removed it doesn't get created.

This is only for the integration tests because in production the priority class is created via ignition.

PR #422, opened 2020-04-15
Only update failed releases if the chart values or version has changed

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/10074

Currently if the release is FAILED we update since the release is not DEPLOYED. However if the values or version have not changed its likely the update will fail again.

This makes the reconciliation process less aggressive but still allows us to deploy fixes.

### DIFF
--- a/service/controller/chart/resource/release/resource.go
+++ b/service/controller/chart/resource/release/resource.go
@@ -183,6 +183,29 @@ func isEmpty(c ReleaseState) bool {
 	return equals(c, ReleaseState{})
 }
 
+func isReleaseFailed(current, desired ReleaseState) bool {
+	if isEmpty(current) {
+		return false
+	}
+
+	// Values have changed so we will try to update the release.
+	if current.ValuesMD5Checksum != desired.ValuesMD5Checksum {
+		return false
+	}
+
+	// Version has changed so we will try to update the release.
+	if current.Version != desired.Version {
+		return false
+	}
+
+	// Release is failed and should not be updated.
+	if current.Status == helmFailedStatus {
+		return true
+	}
+
+	return false
+}
+
 func isReleaseInTransitionState(r ReleaseState) bool {
 	return releaseTransitionStatuses[r.Status]
 }
--- a/service/controller/chart/resource/release/update.go
+++ b/service/controller/chart/resource/release/update.go
@@ -172,7 +172,16 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
 
 	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding out if the %#q release has to be updated", desiredReleaseState.Name))
 
+	// The release is still being updated so we don't update and check again
+	// in the next reconciliation loop.
 	if isReleaseInTransitionState(currentReleaseState) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("the %#q release is in status %#q and cannot be updated", desiredReleaseState.Name, currentReleaseState.Status))
+		return nil, nil
+	}
+
+	// The release is failed and the values and version have not changed. So we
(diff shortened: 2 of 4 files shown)


service/controller/chart/resource/release/update.go
```diff
@@ -172,11 +172,20 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
 
 	r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding out if the %#q release has to be updated", desiredReleaseState.Name))
 
+	// The release is still being updated so we don't update and check again
+	// in the next reconciliation loop.
```
rossf7, line 176: Just adding some comments to give more context.

service/controller/chart/resource/release/update.go
```diff
@@ -172,11 +172,20 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
...
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("the %#q release is in status %#q and cannot be updated", desiredReleaseState.Name, currentReleaseState.Status))
 		return nil, nil
 	}
 
+	// The release is failed and the values and version have not changed. So we
+	// don't update. We will be alerted so we can investigate.
+	if isReleaseFailed(currentReleaseState, desiredReleaseState) {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("the %#q release is in status %#q and cannot be updated", desiredReleaseState.Name, currentReleaseState.Status))
+		return nil, nil
+	}
```
rossf7, line 187: This is the fix.

service/controller/chart/resource/release/resource.go
```diff
@@ -183,6 +183,29 @@ func isEmpty(c ReleaseState) bool {
 	return equals(c, ReleaseState{})
 }
 
+func isReleaseFailed(current, desired ReleaseState) bool {
```
yasn77, line 186: When I initially read this, I had to double back because of the switch between `true` and `false` in the conditionals. Maybe this would be more readable:

```go
func isReleaseFailed(current, desired ReleaseState) bool {
        retVal := false
        
	if ! isEmpty(current) {
	  
          // Values have changed so we will try to update the release.
	  if current.ValuesMD5Checksum == desired.ValuesMD5Checksum {
	  	  retVal = true
	  }

	  // Version has changed so we will try to update the release.
	  if current.Version == desired.Version {
		  retVal = true
	  }

	  // Release is failed and should not be updated.
	  if current.Status == helmFailedStatus {
		  retVal = true
	  }
	
	}

	return retVal
}
```

To make it terser, could just make it one just logical OR conditional... But I think the multiple `if` probably makes it more readable.

rossf7: @yasn77 Go is more verbose than other languages but this is a pretty common pattern.

The problem is this is based on `isReleaseModified` and I don't want to change that as well. With the helm3 branch and flattening the operator I don't want to change more that is required.

@tomahawk28 WDYT?

rossf7:
Tested on `gauss`.

service/controller/chart/resource/release/resource.go
tomahawk28: flapping `true` or `false` is a bit hard to read, of course, @yasn77 is right at this point, maybe we can do it on next PR.

tomahawk28: oh, no, this is a new function, so changing doesn't matter so far.

rossf7: I'll do a follow up and also update isReleaseModified to keep them in sync. I want to get this merged so I can cut a release. As its needed for the KVM 9.0.1 release that is being prepped.

Thanks both for the feedback. ❤️

PR #420, opened 2020-04-14
Remove usage of project.name

rossf7:
Towards giantswarm/giantswarm#7896

PR #423, opened 2020-04-15
helm3: Remove usage of project.name

rossf7:
Same as https://github.com/giantswarm/chart-operator/pull/420 but for helm3 branch.


rossf7:
Going with this. CI failures were due to tarball version and this won't get deployed yet.

PR #424, opened 2020-04-15
Release 0.12.4

rossf7:


pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	name        = "chart-operator"
 	gitSHA      = "n/a"
 	source      = "https://github.com/giantswarm/chart-operator"
-	version     = "1.0.0"
+	version     = "0.12.4"
```
rossf7, line 8: Downgrading version. As we want to use major version 1 to identify Helm 3 releases.

rossf7:
Going with a ping.

PR #425, opened 2020-04-15
Refactor annotations and update state calculation

rossf7:
Follow up to https://github.com/giantswarm/chart-operator/pull/422.

### DIFF
--- a/pkg/annotation/annotation.go
+++ b/pkg/annotation/annotation.go
@@ -14,4 +14,12 @@ const (
 	// DeleteCustomResourceOnly is the name of the annotation that indicates
 	// the custom resource should be deleted without deleting the Helm release.
 	DeleteCustomResourceOnly = "chart-operator.giantswarm.io/delete-custom-resource-only"
+
+	// ForceHelmUpgrade is the name of the annotation that controls whether
+	// force is used when upgrading the Helm release.
+	ForceHelmUpgrade = "chart-operator.giantswarm.io/force-helm-upgrade"
+
+	// ValuesMD5Checksum is the name of the annotation storing an MD5 checksum
+	// of the Helm release values.
+	ValuesMD5Checksum = "chart-operator.giantswarm.io/values-md5-checksum"
 )
--- a/pkg/label/label.go
+++ b/pkg/label/label.go
@@ -8,4 +8,7 @@ const (
 
 	// ManagedBy is set for Kubernetes resources managed by the operator.
 	ManagedBy = "giantswarm.io/managed-by"
+
+	// Version is the version label for chart custom resources.
+	Version = "chart-operator.giantswarm.io/version"
 )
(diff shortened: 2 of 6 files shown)


pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	name        = "chart-operator"
 	gitSHA      = "n/a"
 	source      = "https://github.com/giantswarm/chart-operator"
-	version     = "0.12.4"
+	version     = "0.12.5-dev"
```
rossf7, line 8: Sneaking this in to set the dev version.

pkg/annotation/annotation.go
```diff
@@ -14,4 +14,12 @@ const (
...
 	// the custom resource should be deleted without deleting the Helm release.
 	DeleteCustomResourceOnly = "chart-operator.giantswarm.io/delete-custom-resource-only"
+
+	// ForceHelmUpgrade is the name of the annotation that controls whether
+	// force is used when upgrading the Helm release.
+	ForceHelmUpgrade = "chart-operator.giantswarm.io/force-helm-upgrade"
+
+	// ValuesMD5Checksum is the name of the annotation storing an MD5 checksum
+	// of the Helm release values.
+	ValuesMD5Checksum = "chart-operator.giantswarm.io/values-md5-checksum"
```
rossf7, line 24: Moving these from the key package to the right home.

pkg/label/label.go
```diff
@@ -8,4 +8,7 @@ const (
 
 	// ManagedBy is set for Kubernetes resources managed by the operator.
 	ManagedBy = "giantswarm.io/managed-by"
+
+	// Version is the version label for chart custom resources.
+	Version = "chart-operator.giantswarm.io/version"
```
rossf7, line 13: Same for the version label.

service/controller/chart/resource/release/resource.go
```diff
@@ -184,51 +185,50 @@ func isEmpty(c ReleaseState) bool {
 }
 
 func isReleaseFailed(current, desired ReleaseState) bool {
-	if isEmpty(current) {
-		return false
-	}
+	result := false
```
rossf7, line 188: See https://github.com/giantswarm/chart-operator/pull/422/files#r408751841.

service/controller/chart/resource/release/resource.go
```diff
@@ -183,52 +184,56 @@ func isEmpty(c ReleaseState) bool {
...
-	// Values have changed so we will try to update the release.
-	if current.ValuesMD5Checksum != desired.ValuesMD5Checksum {
-		return false
-	}
+	if !isEmpty(current) {
+		// Values have changed so we should try to update even if the release
+		// is failed.
+		if current.ValuesMD5Checksum != desired.ValuesMD5Checksum {
+			return false
+		}
```
rossf7, line 198: @yasn77 @tomahawk28 This just needed a tweak.

The early return is needed so the log messages are correct. We should log `the release does not have to be updated` rather than `the release is in status DEPLOYED and cannot be updated.`

But it's still a lot cleaner which make me happy. Thx!

rossf7:
Tested on `geckon`.

PR #426, opened 2020-04-15
helm3: Only update failed releases if the chart values or version has changed

rossf7:
Cherry pick of #422 and #425 into helm3 branch.


rossf7:
Going with a ping. CI failures are due to the tarball version. Will get that sorted tomorrow.

PR #429, opened 2020-04-17
Remove e2e-harness

rossf7:
This is needed for the flattening. So that we install chart-operator from the default catalog and not appr.

We already removed e2e-harness from the helm3 branch in https://github.com/giantswarm/chart-operator/pull/376. This gets the master branch in sync.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,8 +4,8 @@ go 1.13
 
 require (
 	github.com/giantswarm/apiextensions v0.2.3
+	github.com/giantswarm/appcatalog v0.1.11
 	github.com/giantswarm/backoff v0.2.0
-	github.com/giantswarm/e2e-harness v0.2.0
 	github.com/giantswarm/e2esetup v0.2.0
 	github.com/giantswarm/exporterkit v0.2.0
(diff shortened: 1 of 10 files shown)


go.mod
```diff
@@ -4,8 +4,8 @@ go 1.13
 
 require (
 	github.com/giantswarm/apiextensions v0.2.3
+	github.com/giantswarm/appcatalog v0.1.11
 	github.com/giantswarm/backoff v0.2.0
-	github.com/giantswarm/e2e-harness v0.2.0
```
rossf7, line 8: 🔥

integration/templates/chart_operator_values.go
```diff
@@ -6,7 +6,7 @@ package templates
 const ChartOperatorValues = `cnr:
   address: http://cnr-server:5000
 clusterDNSIP: 10.96.0.10
-e2e: true
+e2e: false
```
rossf7, line 9: This is so the priority class is not created. Since EnsureTillerInstalled does this.

rossf7:
Going with a ping.

PR #421, opened 2020-04-14
Deploy as an app CR in control planes

rossf7:
Towards giantswarm/giantswarm#7896

This completes the flattening code changes. Needs https://github.com/giantswarm/app-operator/pull/255 to be merged first. 

Once this is merged we'll create a 0.13.0 release. We want to use major version 1 to distinguish chart-operator CRs using Helm 3.

PR #433, opened 2020-04-22
Delete unused helm chart

rossf7:
Follow up to https://github.com/giantswarm/chart-operator/pull/421

We can now delete the helm chart that was used by draughtsman.


rossf7:
Going with a ping.

PR #437, opened 2020-04-23
Use architect/go-build

rossf7:


service/controller/chart/resource/release/desired.go
```diff
@@ -46,7 +46,9 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 			return nil, microerror.Mask(err)
 		}
 
-		valuesMD5Checksum = fmt.Sprintf("%x", md5.Sum(valuesYAML))
+		// MD5 is only used for comparison but we need to turn off gosec or
+		// linting errors will occur.
+		valuesMD5Checksum = fmt.Sprintf("%x", md5.Sum(valuesYAML)) // #nosec
```
rossf7, line 51: We use this MD5 checksum annotation to check if the helm release values have changed. This is because we can't rely on Helm to return the same values we provide.

I think MD5 is safe to use for this but FYI @stone-z 

```
apiVersion: application.giantswarm.io/v1alpha1
kind: Chart
metadata:
  annotations:
    chart-operator.giantswarm.io/values-md5-checksum: 124a7cadff8175fd68d0f8933b7fb161
```

service/controller/chart/resource/status/create.go
```diff
@@ -91,7 +91,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		AppVersion: releaseHistory.AppVersion,
 		Reason:     reason,
 		Release: v1alpha1.ChartStatusRelease{
-			LastDeployed: v1alpha1.DeepCopyTime{releaseHistory.LastDeployed},
+			LastDeployed: metav1.NewTime(releaseHistory.LastDeployed),
```
rossf7, line 94: <3

.circleci/config.yml
```diff
@@ -1,29 +1,14 @@
...
+  architect: giantswarm/architect@0.8.12
 
 workflows:
   build:
     jobs:
-      - build:
+      - architect/go-build:
+          name: go-build
+          binary: chart-operator
+          resource_class: large
```
kopiczko, line 11: Isn't medium enough here?

```suggestion
```

rossf7: Yes probably but I'd like to see the speed difference. With large it takes 01:26.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
+- Use go build from architect orb.
```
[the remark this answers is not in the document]
rossf7: Ack. Removed.

.circleci/config.yml
rossf7: Time is the same. I'll also downgrade app-operator. 👍

service/controller/chart/resource/release/desired.go
stone-z: If it's for change detection and not protecting data, you're good to go :+1:

service/controller/chart/resource/chartmigration/error_internal.go
```diff
@@ -37,7 +37,7 @@ func isChartConfigNotAvailable(err error) bool {
 		}
 	}
 
-	if c == chartConfigNotAvailableError {
+	if c == chartConfigNotAvailableError { //nolint:gosimple
```
stone-z, line 40: Not a blocker, but since this returns right after anyway, why suppress the gosimple recommendation?

service/service.go
```diff
@@ -184,7 +184,7 @@ func New(config Config) (*Service, error) {
 // Boot starts top level service implementation.
 func (s *Service) Boot(ctx context.Context) {
 	s.bootOnce.Do(func() {
-		go s.operatorCollector.Boot(ctx)
+		go s.operatorCollector.Boot(ctx) // nolint:errcheck
```
stone-z, line 187: There's a sample fix for this errcheck in kvm-operator if you like

rossf7: Nice thanks I've updated to use that.

service/controller/chart/resource/chartmigration/error_internal.go
rossf7: Its not our usual code style and looked a bit strange to me. 

So I checked the errors library. This is the same thing but for the chartconfig CRD. So I'd prefer to leave this one as is.

https://github.com/giantswarm/errors/blob/d79b72c74790d1d964ecc87013a37fdbd1f86b0b/tenant/error.go#L59-L61

PR #443, opened 2020-04-28
helm3 changes

tomahawk28:
This is new branch for seeking the amount of changes from `helm3` branch. 
If we agree, we kept this and delete `helm3` branch.


tomahawk28:
Closing now, we can come back later

PR #449, opened 2020-05-06
helm3; update to 1.0.0

tomahawk28:
For helm3 supported version operators; we should know that having a major version as 1 means it could process helm release as helm3 protocol.

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -5,7 +5,7 @@ var (
 	gitSHA      = "n/a"
 	name        = "chart-operator"
 	source      = "https://github.com/giantswarm/chart-operator"
-	version     = "0.13.1-dev"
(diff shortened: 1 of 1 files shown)


tomahawk28:
> Looks fine but didn't we have problems with 1.X before?
> 
> Worst case can we use a hardcoded version and replace it once we're ready to deploy?

I think that happened because of below line in app-operator. 

https://github.com/giantswarm/app-operator/blob/c0ebff892744814fc435006bedd0b19c5433046c/integration/test/app/basic/simple_test.go#L48

but this is just an app version update and it merely lives in the test catalog so I think it's fine.

rossf7:
OK cool we need to use v1.0.0 for the helm 3 release. So no blocker if this helps. 👍

PR #460, opened 2020-05-11
Update go modules

rossf7:


rossf7:
Only dependencies being updated. CI green. Going with a ping.

PR #461, opened 2020-05-11
helm3: Update go modules

rossf7:


rossf7:
Same as https://github.com/giantswarm/chart-operator/pull/460 to keep helm3 branch in sync. Going with a ping.

PR #466, opened 2020-05-13
Fixing empty lastDeployed when release deployment is not successful 

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/10644

Filling empty deployment timestamp in case of deployment failure.


tomahawk28:
Not sure why migration logic e2e failed but seems it's not related with this.

tomahawk28:
going with this, and would apply same for helm3 as cherry-pick

PR #467, opened 2020-05-13
Helm3: fixing empty lastdeployed 

tomahawk28:
Same as https://github.com/giantswarm/chart-operator/pull/466 but only for helm3.


tomahawk28:
Same as helm3, would merge it after CI claims green.

tomahawk28:
going with a ping

PR #471, opened 2020-05-18
Apply helm3 changes

tomahawk28:

Applying helmclient 1.0.0 and adding a changelog.


service/collector/helm_v2_release.go
```diff
@@ -0,0 +1,100 @@
...
+	return h, nil
+}
+
+func (h *HelmV2Release) Collect(ch chan<- prometheus.Metric) error {
+	lo := metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("%s=%s", "OWNER", "TILLER"),
+	}
+
+	// Check whether helm 2 release configMaps still exist.
+	cms, err := h.k8sClient.CoreV1().ConfigMaps(h.tillerNamespace).List(lo)
```
tomahawk28, line 65: This is actually good metrics, so we can keep working until only draughtsman release left.

PR #472, opened 2020-05-18
Release v1.0.0

tomahawk28:

Fixing project.go version and changelog.

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -5,7 +5,7 @@ var (
 	gitSHA      = "n/a"
 	name        = "chart-operator"
 	source      = "https://github.com/giantswarm/chart-operator"
-	version     = "1.0.0-dev"
+	version     = "1.0.0"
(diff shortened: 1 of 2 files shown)


pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA      = "n/a"
 	name        = "chart-operator"
 	source      = "https://github.com/giantswarm/chart-operator"
-	version     = "1.0.0-dev"
+	version     = "1.0.0"
```
tomahawk28, line 8: I will rollback it after pushing tags.

rossf7: Thanks. Minor but we should set the version to `1.0.1-dev` so it isn't a rollback.

tomahawk28: Yes, I will upgrade it to 1.0.1-dev after tags.

PR #474, opened 2020-05-18
version upgrade to 1.0.1-dev

tomahawk28:

Since 1.0.0 chart-operator is released, we upgrade the `project.go` file accordingly.


tomahawk28:
going with a ping

PR #480, opened 2020-05-25
Avoid validation failure against not-installed releases

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/10644

Filling `lastDeployed` for blank date in case of release had been failed to install.

### DIFF
--- a/service/controller/chart/resource/status/create.go
+++ b/service/controller/chart/resource/status/create.go
@@ -3,6 +3,7 @@ package status
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/helmclient"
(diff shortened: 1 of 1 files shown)


tomahawk28:
Checked validation passed. 
```
  status:
    appVersion: ""
    reason: FFFF
    release:
      lastDeployed: "0000-01-01T00:00:00Z"
      revision: 0
      status: FFFF
    version: ""
```

service/controller/chart/resource/status/create.go
```diff
@@ -13,6 +14,10 @@ import (
 	"github.com/giantswarm/chart-operator/service/controller/chart/key"
 )
 
+var (
+	blankDate = time.Date(0, time.January, 1, 0, 0, 0, 0, time.UTC)
```
rossf7, line 18: For me this is a sign that the openapi rules in the app CRD are wrong.

If the helm release fails to deploy then there is never a date. What about fixing this in apiextensions instead?

tomahawk28: Good catch, That could be a more plausible option. I would take a look into apiextetions repo.

tomahawk28:
Closing in favor of https://github.com/giantswarm/apiextensions/pull/483

PR #483, opened 2020-05-26
Checking validation failure on release manifest

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/11167

Canceling release migration in case of validation failure.


service/controller/chart/resource/release/create.go
```diff
@@ -104,7 +104,16 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 			return nil
 		}
 
-		if err != nil {
+		if helmclient.IsInvalidManifest(err) {
+			reason := err.Error()
+			reason = fmt.Sprintf("helm validation error: (%s)", reason)
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("helm release %#q failed, %s", releaseState.Name, reason))
```
tomahawk28, line 110: I would like to put this as a one-line debugging code.

PR #487, opened 2020-05-27
Canceling release resource if helm 2 release is not deleted

tomahawk28:
Following https://github.com/giantswarm/giantswarm/issues/10973#issuecomment-634473564 

chart-operator is creating helm 3 release even before app-operator starts cordon all charts and start the migration. 

This patch would prevent that early behavior.

### DIFF
--- a/service/controller/chart/resource/release/create_test.go
+++ b/service/controller/chart/resource/release/create_test.go
@@ -75,6 +75,8 @@ func Test_Resource_Release_newCreate(t *testing.T) {
 			HelmClient: helmclienttest.New(helmclienttest.Config{}),
 			K8sClient:  k8sfake.NewSimpleClientset(),
 			Logger:     microloggertest.New(),
+
+			TillerNamespace: "giantswarm",
 		}
 
 		newResource, err = New(c)
--- a/service/controller/chart/resource/release/current.go
+++ b/service/controller/chart/resource/release/current.go
@@ -36,7 +36,7 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	}
 
 	if hasConfigmap {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("release %#q has been not migrated from helm 2", key.ReleaseName(cr)))
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("release %#q has not been migrated from helm 2", key.ReleaseName(cr)))
 		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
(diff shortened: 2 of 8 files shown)


service/controller/chart/resource/release/resource.go
```diff
@@ -89,6 +103,20 @@ func (r *Resource) Name() string {
 	return Name
 }
 
+func (r *Resource) findHelmV2ConfigMaps(ctx context.Context, releaseName string) (bool, error) {
+	lo := metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("%s=%s,%s=%s", "NAME", releaseName, "OWNER", "TILLER"),
+	}
+
+	// Check whether it keep helm2 release configMaps
```
rossf7, line 111:
```suggestion
	// Check whether there are still helm2 release configmaps.
```

service/controller/chart/resource/release/current.go
```diff
@@ -30,6 +30,18 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		return nil, nil
 	}
 
+	hasConfigmap, err := r.findHelmV2ConfigMaps(ctx, key.ReleaseName(cr))
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	if hasConfigmap {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("release %#q has been not migrated from helm 2", key.ReleaseName(cr)))
```
rossf7, line 39:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("release %#q has not been migrated from helm 2", key.ReleaseName(cr)))
```

service/controller/chart/resource/release/create_test.go
```diff
@@ -75,6 +75,8 @@ func Test_Resource_Release_newCreate(t *testing.T) {
 			HelmClient: helmclienttest.New(helmclienttest.Config{}),
 			K8sClient:  k8sfake.NewSimpleClientset(),
 			Logger:     microloggertest.New(),
+
+			TillerNamespace: "giantswarm",
```
rossf7, line 79: Not for now but I'm wondering if we should drop the unit tests apart from the desired state.

I think we already did that in app-operator. I'd rather extend our existing integration tests or add new ones. WDYT?

service/controller/chart/resource/release/current.go
tomahawk28: Thanks for helping 🇬🇧 :)

service/controller/chart/resource/release/create_test.go
tomahawk28: Good call, I would delete them in next PR.

PR #495, opened 2020-06-03
Calculate md5 from struct

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/11337

Fixing helm 3 supported version, and later I would apply same to 0.13.x.

### DIFF
--- a/service/controller/chart/resource/release/desired.go
+++ b/service/controller/chart/resource/release/desired.go
@@ -51,7 +51,9 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 
 		// MD5 is only used for comparison but we need to turn off gosec or
 		// linting errors will occur.
-		valuesMD5Checksum = fmt.Sprintf("%x", md5.Sum(valuesYAML)) // #nosec
+		h := md5.New()
+		h.Write([]byte(fmt.Sprintf("%v", values)))
+		valuesMD5Checksum = fmt.Sprintf("%x", h.Sum(nil))
 	}
 
 	releaseState := &ReleaseState{
--- a/service/controller/chart/resource/release/desired.go
+++ b/service/controller/chart/resource/release/desired.go
@@ -41,8 +41,12 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 	if len(values) > 0 {
 		// MD5 is only used for comparison but we need to turn off gosec or
 		// linting errors will occur.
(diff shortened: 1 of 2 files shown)


service/controller/chart/resource/release/desired.go
```diff
@@ -37,21 +36,18 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
 
 	if len(values) > 0 {
-		// We serialize the values to YAML so we can generate the MD5 checksum.
-		// We use this for comparison because Helm may modify the values we get
-		// back from the Helm client.
-		valuesYAML, err = yaml.Marshal(values)
+		// MD5 is only used for comparison but we need to turn off gosec or
+		// linting errors will occur.
+		h := md5.New() // #nosec
+		_, err := h.Write([]byte(fmt.Sprintf("%v", values)))
```
tomahawk28, line 45: @oponder I heard you are interested in this part, getting hash value from the struct.

service/controller/chart/resource/release/desired_test.go
```diff
@@ -224,7 +224,7 @@ func Test_DesiredState(t *testing.T) {
 			expectedState: ReleaseState{
 				Name:              "chart-operator-chart",
 				Status:            helmclient.StatusDeployed,
-				ValuesMD5Checksum: "8ccfa2ed7f5cb9a125b5f53254c296a8",
+				ValuesMD5Checksum: "2187a8fce91c3765a74d462062af7526",
```
tomahawk28, line 227: This means after we deploy this in CP, it would likely upgrade every charts, but this is needed.

rossf7: Thanks for the heads up but this is fine. We can verify once we've deployed that it only updates once.

rossf7:
@tomahawk28 I like the direction but please verify it works with the problem data. This PR may help https://github.com/giantswarm/installations/pull/1278.

service/controller/chart/resource/release/desired_test.go
tomahawk28: Since all CPs still have chart-operator with helm 2 support, I think it makes sense to test new branch based on `v0.13.x`. WDYT?

rossf7: Yes good point. Better to test with Helm 2 first.

tomahawk28:
@giantswarm/team-batman-engineers Can we go with this?

PR #497, opened 2020-06-04
Disable force upgrade from helm 3

tomahawk28:
Disabling force-upgrade from helm3 to mitigate errors from helmclient. 

For more about it, please read; https://github.com/helm/helm/issues/7956#issuecomment-632274443.

### DIFF
--- a/service/controller/chart/resource/release/update.go
+++ b/service/controller/chart/resource/release/update.go
@@ -32,6 +32,10 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
 	upgradeForce := key.HasForceUpgradeAnnotation(cr)
 
 	if releaseState.Name != "" {
+		// TODO: Disabling upgrade-force from chart-operator 1.0.2
+		//
+		//	See https://github.com/giantswarm/giantswarm/issues/11376
+		//
 		if upgradeForce {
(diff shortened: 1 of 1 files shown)


service/controller/chart/resource/release/update.go
```diff
@@ -32,8 +32,13 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
...
 
 	if releaseState.Name != "" {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("updating release %#q with force == %t", releaseState.Name, upgradeForce))
-
+		// TODO: Disabling upgrade-force from chart-operator 1.0.2
+		//
+		//	See https://github.com/giantswarm/giantswarm/issues/11376
+		//
+		if upgradeForce {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("disabling helm upgrade-force on release %#q", releaseState.Name))
```
rossf7, line 40:
```suggestion
			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("helm upgrade force is disabled for %#q", releaseState.Name))
```

What about this?

tomahawk28: Nice, Thanks for notice.

PR #501, opened 2020-06-11
Upgrade Tiller image to 2.16.8

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/11510

⚠️ We must deploy app-operator https://github.com/giantswarm/app-operator/pull/367 first, 
otherwise two operators would fight over image version of tiller deploy. 

Upgrading tiller to 2.16.8 and add new flag `TillerImageName` for more usability.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/giantswarm/chart-operator
 
-go 1.13
+go 1.14
 
 require (
 	github.com/giantswarm/apiextensions v0.2.3
--- a/helm/chart-operator/templates/configmap.yaml
+++ b/helm/chart-operator/templates/configmap.yaml
@@ -16,11 +16,11 @@ data:
     service:
       helm:
         http:
-          {{ if (.Values.Installation) }}
+          {{- if (.Values.Installation) }}
           clientTimeout: '{{ .Values.Installation.V1.Helm.HTTP.ClientTimeout }}'
-          {{ else }}
+          {{- else }}
           clientTimeout: '{{ .Values.helm.http.clientTimeout }}'
-          {{ end }}
+          {{- end }}
         tillerImageName:  '{{ .Values.tiller.image }}'
         tillerNamespace:  '{{ .Values.tiller.namespace }}'
       image:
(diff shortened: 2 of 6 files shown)


helm/chart-operator/templates/configmap.yaml
```diff
@@ -16,11 +16,12 @@ data:
...
         http:
-          {{ if (.Values.Installation) }}
+          {{- if (.Values.Installation) }}
           clientTimeout: '{{ .Values.Installation.V1.Helm.HTTP.ClientTimeout }}'
-          {{ else }}
+          {{- else }}
           clientTimeout: '{{ .Values.helm.http.clientTimeout }}'
-          {{ end }}
+          {{- end }}
+        tillerImageName:  '{{ .Values.tiller.image }}'
```
tomahawk28, line 24: Before: 
```
service:
  helm:
    http:

      clientTimeout: '5s'

    tillerImageName:  'giantswarm/tiller:v2.16.8'
    tillerNamespace:  'giantswarm'
```

After:
```
service:
  helm:
    http:
      clientTimeout: '5s'
    tillerImageName:  'giantswarm/tiller:v2.16.8'
    tillerNamespace:  'giantswarm'
```

helm/chart-operator/values.yaml
```diff
@@ -45,6 +45,7 @@ resource:
     name: '{{ .Release.Name | replace "." "-" | trunc 47 }}-psp'
 
 tiller:
+  image: "giantswarm/tiller:v2.16.8"
```
tomahawk28, line 48: Later when we would like to update a tiller after this security incident, we can simple override this value in default-catalog configmap.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/giantswarm/chart-operator
 
-go 1.13
+go 1.14
```
tomahawk28, line 3: architect-orb is already 0.8.8+ which means it would built by go 1.14 when we build images, but for safety reason I would like to update it as 1.14.

PR #508, opened 2020-06-22
Collect helm release namespace

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/11651


go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/giantswarm/backoff v0.2.0
 	github.com/giantswarm/e2esetup v0.2.0
 	github.com/giantswarm/exporterkit v0.2.0
-	github.com/giantswarm/helmclient v0.2.2
+	github.com/giantswarm/helmclient v0.2.4-0.20200622064529-892b681a480a
```
tomahawk28, line 11: TODO: revert when new helmclient is released

service/collector/collector.go
```diff
@@ -1,7 +1,8 @@
 package collector
 
 const (
-	Namespace = "chart_operator"
+	gaugeValue float64 = 1
+	namespace          = "chart_operator"
```
rossf7, line 5: <3

service/collector/orphan_configmap.go
```diff
@@ -17,7 +17,7 @@ import (
 
 var (
 	orphanConfigMapDesc *prometheus.Desc = prometheus.NewDesc(
-		prometheus.BuildFQName(Namespace, "configmap", "orphan"),
+		prometheus.BuildFQName(namespace, "configmap", "orphan"),
```
rossf7, line 20: For my interest does this mean we also need to change g8s-prometheus?

tomahawk28: No, I changed the name of`Namespace` variable into `namespace`, so no need to change anything from g8s-prometheus. Sorry for the confusion.

service/collector/namespace_inconsistency.go
```diff
@@ -0,0 +1,103 @@
...
+	var value float64
+	for _, chart := range charts.Items {
+		content, err := n.helmClient.GetReleaseContent(ctx, chart.Spec.Name)
+		if err != nil {
+			n.logger.Log("level", "error", "message", "failed to collect namespace consistency", "stack", fmt.Sprintf("%#v", err))
+			continue
+		}
+
+		if key.Namespace(chart) != content.Namespace {
+			value = 1
```
tomahawk28, line 82: Set gauge as 1 only in case of inconsistency found.

service/collector/collector.go
```diff
@@ -1,13 +1,11 @@
 package collector
 
 const (
-	gaugeValue float64 = 1
-	Namespace          = "chart_operator"
+	namespace = "chart_operator"
```
tomahawk28, line 4: `namespace` only belongs to collector itself, so changed it to small-case naming.

service/collector/namespace_inconsistency.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/chart-operator/service/controller/chart/key"
+)
+
+var (
+	namespaceInconsistency = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, "", "namespace_inconsistency"),
+		"namespace is consistent with chart CR spec.",
```
rossf7, line 20:
```suggestion
		"namespace is inconsistent with chart CR spec.",
```
Should this match the metric name?


Naming is hard here but I'm nor sure about inconsistency. Could we use mismatch? As it is also shorter.

But I'm also not sure that is better. So go with what you think is best.

service/collector/orphan_configmap.go
rossf7: No problem and I could have got that. Sorry it took so long to get back to this.

service/collector/namespace_inconsistency.go
tomahawk28: `mismatch` is a lot better I think, I will change the name to that. Thanks for the input!
`inconsistency` felt like too much scientific term 😆 .

helm/chart-operator/values.yaml
```diff
@@ -45,7 +45,6 @@ resource:
     name: '{{ .Release.Name | replace "." "-" | trunc 47 }}-psp'
 
 tiller:
-  image: "giantswarm/tiller:v2.16.8"
```
tomahawk28, line 48: This is the leftover from last Helm security incident. We need to remove this before rolling out a new release.

rossf7: Good catch, thanks!

tomahawk28:
Tested in gorgoth. 
```
chart_operator_namespace_mismatch{chart_name="releases-kvm-unique",namespace="giantswarm",placed_namespace="giantswarm"} 0
chart_operator_namespace_mismatch{chart_name="tokend-unique",namespace="giantswarm",placed_namespace="giantswarm"} 0
chart_operator_namespace_mismatch{chart_name="userd-unique",namespace="giantswarm",placed_namespace="giantswarm"} 0
```

PR #516, opened 2020-07-01
Update chart CR status in case of migration not had been done

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/11907

Updating chart CR status in case of migration had been not performed yet.

### DIFF
--- a/service/controller/chart/resource/release/current.go
+++ b/service/controller/chart/resource/release/current.go
@@ -32,6 +32,8 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 
 	hasConfigmap, err := r.findHelmV2ConfigMaps(ctx, key.ReleaseName(cr))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


service/controller/chart/resource/release/current.go
```diff
@@ -32,6 +32,8 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 
 	hasConfigmap, err := r.findHelmV2ConfigMaps(ctx, key.ReleaseName(cr))
 	if err != nil {
+		reason := fmt.Sprintf("release %#q didn't migrate to helm 3", key.ReleaseName(cr))
+		addStatusToContext(cc, reason, releaseNotInstalledStatus)
```
tomahawk28, line 36: Only those changes are needed.

tomahawk28:
@giantswarm/team-batman-engineers Can anyone look into this before we release new chart-operator?

PR #518, opened 2020-07-06
Enable pushing to app collections and update orb

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/10973

Now all CPs are on Helm 3 we should enable pushing to the collections again.


rossf7:
#TooSoon converted back to draft.

PR #519, opened 2020-07-06
Make k8s wait timeout configurable

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/11969


helm/chart-operator/values.yaml
```diff
@@ -9,6 +9,8 @@ externalDNSIP: 8.8.8.8
 helm:
   http:
     clientTimeout: "5s"
+  kubernetes:
+    waitTimeout: "10s"
```
rossf7, line 13: Increasing the wait timeout from `3s` to `10s`. Except in China where we will use `30s`.

service/controller/chart/resource/release/create.go
```diff
@@ -95,7 +94,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 	select {
 	case <-ch:
 		// Fall through.
-	case <-time.After(3 * time.Second):
+	case <-time.After(r.k8sWaitTimeout):
```
rossf7, line 97: This is the change.

service/controller/chart/resource/release/update.go
```diff
@@ -108,7 +107,7 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
 	select {
 	case <-ch:
 		// Fall through.
-	case <-time.After(3 * time.Second):
+	case <-time.After(r.k8sWaitTimeout):
```
rossf7, line 110: Same for updating releases.

helm/chart-operator/values.yaml
tomahawk28: Can we test this value with a prometheus-operator? I think it's valid to check with the chart which takes the longest hours.

rossf7: Good call. I'll do that later on another CP. I've been testing on giraffe but its so slow its unusable. 

I don't want us to block because prometheus-operator takes too long. If this doesn't work then we may need to fix the prometheus-operator chart and try and make it faster.

rossf7: I tested with prometheus-operator on giraffe and it worked fine. But I want to test more on giraffe in the morning.

rossf7:
Tested on `giraffe`. Going with this.

PR #520, opened 2020-07-06
Update go modules and use latest architect

rossf7:
Gettings deps updated and fixing CI.


rossf7:

PR #521, opened 2020-07-06
Add workflows

rossf7:
Adding workflows generated by `devctl gen workflows`.


tomahawk28:
Can you add README.md about how to release new version?

rossf7:
> Can you add README.md about how to release new version?

I want our readme to be the same as all other repos. So if something needs to be in the README it should come from biscuit.

Please watch the video in https://gigantic.slack.com/archives/C04TGHDEF/p1593791537176700 when you have time. Its only 11 mins.

PR #522, opened 2020-07-07
Set release revision in CR status

rossf7:
Towards giantswarm/giantswarm#11969

```bash
kg get chart prometheus-operator-app-unique -o yaml | yq .status.release
{
  "lastDeployed": "2020-07-07T08:55:00Z",
  "revision": 2,
  "status": "deployed"
}
```

### DIFF
--- a/service/controller/chart/resource/status/create.go
+++ b/service/controller/chart/resource/status/create.go
@@ -83,6 +83,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		Reason:     reason,
 		Release: v1alpha1.ChartStatusRelease{
 			LastDeployed: metav1.Time{Time: releaseContent.LastDeployed},
+			Revision:     releaseContent.Revision,
 			Status:       status,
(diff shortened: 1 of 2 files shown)


service/controller/chart/resource/status/create.go
```diff
@@ -83,6 +83,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		Reason:     reason,
 		Release: v1alpha1.ChartStatusRelease{
 			LastDeployed: metav1.Time{Time: releaseContent.LastDeployed},
+			Revision:     releaseContent.Revision,
```
rossf7, line 86: I remember what happened now. With Helm 2 we needed to call both `GetReleaseContent` and `GetReleaseHistory`. 

With Helm 3 we only need to call `GetReleaseContent` but I missed setting this field.

rossf7:
Tested on `giraffe`.

rossf7:
Minor change. Going with a ping.

PR #524, opened 2020-07-07
Fix missing wiring for wait timeout

rossf7:
Towards giantswarm/giantswarm#11969

I missed the setting from the resource set config. 🤦‍♂️


rossf7:
Tested on `godsmack`.

rossf7:
Minor change. Going with a ping.

PR #527, opened 2020-07-08
Fix changelog links

rossf7:
For the new release automation its expecting 3 dots not 2 in changelog links.

https://github.com/giantswarm/architect/blob/b85a96fa76e666da98c0df9f68e874c9a8a09896/cmd/preparerelease/internal/modifier.go#L73-L77


rossf7:
Going with a ping.

PR #530, opened 2020-07-08
Fix up README

rossf7:
At least this README is better than app-operator. 😄


rossf7:
Minor changes for consistency. Going with a ping.

PR #533, opened 2020-07-10
re-enable app-collection push

tomahawk28:


.circleci/config.yml
```diff
@@ -65,48 +65,47 @@ workflows:
...
-      #     app_collection_repo: "aws-app-collection"
-      #     unique: "true"
-      #     requires:
-      #       - push-chart-operator-to-control-plane-app-catalog
-      #     filters:
-      #       branches:
-      #         ignore: /.*/
-      #       tags:
-      #         only: /^v.*/
+      - architect/push-to-app-collection:
```
rossf7, line 68: Can you either add the same comment as app-operator, or remove the comment? 

I don't care which. I just care that we are consistent. 😉  Thx!

tomahawk28: I will delete comments in app-operator, thx!

PR #540, opened 2020-07-14
 Add codeowners

rossf7:


.github/dependabot.yml
```diff
@@ -10,8 +10,6 @@ updates:
     interval: daily
     time: "04:00"
   open-pull-requests-limit: 10
-  reviewers:
-  - giantswarm/team-batman-engineers
```
rossf7, line 14: We don't need this as dependabot will use CODEOWNERS.

rossf7:
Going with a ping.

PR #543, opened 2020-07-20
Decrease resources requests

corest:
Towards https://github.com/giantswarm/giantswarm/issues/12243

Here I took the busiest asgard cluster and resources usage for chart-operator is quite low

fb1af03c-8a20-44a8-8a06-b1b38f6d143a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
+- Decrease CPU and memory requests.
+
 ## [1.0.5] - 2020-07-15
 
 ### Changed
--- a/helm/chart-operator/templates/deployment.yaml
+++ b/helm/chart-operator/templates/deployment.yaml
@@ -104,7 +104,7 @@ spec:
           timeoutSeconds: 1
         resources:
           requests:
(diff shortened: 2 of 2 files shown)


corest:
@giantswarm/team-batman-engineers does the test expected to pass?

tomahawk28:
Normally It should pass but it doesn't look affected by your changes.

rossf7:
@corest Did you check the resource usage in CP as well?

I had a problem earlier in `ginger` where the pod OOMed when this was deployed. In CP there are more chart CRs and the values YAML is a lot bigger.

I think we need to increase the resources to support CPs. Or override in the CP via the draughtsman values. WDYT?

tomahawk28:
From what I saw in prometheus, chart-operator nearly use 160M all the times, so I think the monitoring data from grafana have a validation issue. 
image

https://prometheus.g8s.archon.ore.aws.k8s.3stripes.net/graph?g0.range_input=6h&g0.expr=sum%20by%20(pod)%20(container_memory_working_set_bytes%7Bpod%3D~%22chart-operator-unique-.*%22%7D)&g0.tab=0&g1.range_input=1h&g1.expr=&g1.tab=1

I think we are calculating average memory usage including chart-operator from TCs, too. 

We will revert this PR otherwise we are getting OOM from chart-operator.

PR #544, opened 2020-07-20
Updated workflows

rossf7:
`devctl gen workflows --flavour operator`


rossf7:
Build is green. That's enough in this case.

PR #552, opened 2020-07-24
Revert decreasing resource size

tomahawk28:
Reverting https://github.com/giantswarm/chart-operator/pull/543


tomahawk28:
I wrote that reason behind this revert in here; https://github.com/giantswarm/chart-operator/pull/543#issuecomment-663394223 Better to put in here earlier!

PR #556, opened 2020-07-24
Apply K8S 1.18

tomahawk28:
Towards https://github.com/giantswarm/giantswarm/issues/12716


.circleci/config.yml
```diff
@@ -114,11 +114,3 @@ workflows:
           requires:
           - go-build
           - push-chart-operator-to-default-app-catalog
-      
-      - architect/integration-test:
-          context: architect
-          name: migration-integration-test
```
tomahawk28, line 120: Since we could not ensure `chartconfig` CRD using k8sclient/v4, we are deleting this integration test.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+## Changed
+
+- Updated Kubernetes dependencies to v1.18.5.
```
rossf7, line 12:
```suggestion
- Updated backward incompatible Kubernetes dependencies to v1.18.5.
```

Let's use this in both operators.

.circleci/config.yml
rossf7: This migration is nearly done so I think this is fine.

service/collector/helm_v2_release.go
```diff
@@ -62,7 +63,7 @@ func (h *HelmV2Release) Collect(ch chan<- prometheus.Metric) error {
 	}
 
 	// Check whether helm 2 release configMaps still exist.
-	cms, err := h.k8sClient.CoreV1().ConfigMaps(h.tillerNamespace).List(lo)
+	cms, err := h.k8sClient.CoreV1().ConfigMaps(h.tillerNamespace).List(context.Background(), lo)
```
rossf7, line 66: Let's define the context at the start of the method. We might need to reuse it.

tomahawk28:
I will deploy this into `ginger` today and see it until tomorrow morning.

tomahawk28:
No problem since last night deployment. going to merge this 👍

PR #569, opened 2020-08-12
Quit reconciliation right after rollback

tomahawk28:
Some helm releases got updated throttle if it stuck in `pending-upgrade` or `pending-rollback`. 
this is because the chart-operator has a bug to clear the annotation which related to the limit of retries. 

This PR would quit the reconciliation right after rollback so it would not clear the annotation just like now.

### DIFF
--- a/service/controller/chart/resource/release/update.go
+++ b/service/controller/chart/resource/release/update.go
@@ -229,6 +229,8 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
 			if err != nil {
 				return nil, microerror.Mask(err)
 			}
+			// no-op after rollback
+			return nil, nil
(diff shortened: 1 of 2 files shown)


service/controller/chart/resource/release/update.go
```diff
@@ -229,6 +229,8 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
 			if err != nil {
 				return nil, microerror.Mask(err)
 			}
+			// no-op after rollback
```
tomahawk28, line 232: if we don't quit in here, it actually reach the point of https://github.com/giantswarm/chart-operator/pull/569/files#diff-28b21de142787ce2c87ef523fff62312L249

PR #577, opened 2020-08-18
Namespace resource

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/11748 
From helm3, it stops supporting to create new namespace from release creations. Therefore we need to create the namespace from the chart-operator if that namespace is not present in tenant clusters.

### DIFF
--- a/service/controller/chart/resource/namespace/create.go
+++ b/service/controller/chart/resource/namespace/create.go
@@ -49,7 +49,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	}
 
 	if apierrors.IsAlreadyExists(err) {
-		// fall through
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("already created namespace %#q", key.Namespace(cr)))
 	} else if err != nil {
 		return microerror.Mask(err)
 	}
--- a/service/controller/chart/resource/namespace/create.go
+++ b/service/controller/chart/resource/namespace/create.go
@@ -1,0 +1,60 @@
+package namespace
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/giantswarm/apiextensions/v2/pkg/label"
+	"github.com/giantswarm/microerror"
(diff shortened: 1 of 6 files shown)


service/controller/chart/resource/namespace/create.go
```diff
@@ -0,0 +1,60 @@
...
+	cr, err := key.ToCustomResource(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: key.Namespace(cr),
+			Labels: map[string]string{
+				label.ManagedBy: project.Name(),
```
tomahawk28, line 28: I'm not sure which labels are necessary for here as additional.

service/controller/chart/resource/namespace/create.go
```diff
@@ -0,0 +1,60 @@
...
+	case <-ch:
+		// Fall through.
+	case <-time.After(r.k8sWaitTimeout):
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("timeout creating namespace %#q", key.Namespace(cr)))
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+		return nil
+	}
+
+	if apierrors.IsAlreadyExists(err) {
+		// fall through
```
rossf7, line 52:
```suggestion
			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("already created namespace %#q", key.Namespace(cr)))
			return nil
```

What about logging the namespace already exists? Otherwise it might look like we're creating the namespace each time.

service/controller/chart/resource/namespace/create.go
rossf7: I think just setting the managed-by label makes sense. We can add more later if necessary.

service/controller/chart/resource/namespace/create.go
tomahawk28: Good idea, I will add the following logs.

PR #592, opened 2020-09-04
Remove memory limits from deployment

rossf7:
chart-operator has started OOMing in a few control planes. Removing memory limits so it can run.

https://gigantic.slack.com/archives/C03CPNRTS/p1599222623248300


sslavic (approved):
LGTM

PR #595, opened 2020-09-14
Fix namespace mismatch metric

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/13122

### DIFF
--- a/service/collector/namespace_mismatch.go
+++ b/service/collector/namespace_mismatch.go
@@ -70,11 +70,12 @@ func (n *NamespaceMismatch) Collect(ch chan<- prometheus.Metric) error {
 		return microerror.Mask(err)
 	}
 
-	var value float64
 	for _, chart := range charts.Items {
+		var value float64 = 0
+
 		content, err := n.helmClient.GetReleaseContent(ctx, chart.Spec.Name)
 		if err != nil {
-			n.logger.Log("level", "error", "message", "failed to collect namespace consistency", "stack", fmt.Sprintf("%#v", err))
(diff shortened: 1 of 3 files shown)


service/collector/namespace_mismatch.go
```diff
@@ -70,11 +70,14 @@ func (n *NamespaceMismatch) Collect(ch chan<- prometheus.Metric) error {
...
 
-	var value float64
 	for _, chart := range charts.Items {
+		var value float64 = 0
+
 		content, err := n.helmClient.GetReleaseContent(ctx, chart.Spec.Name)
-		if err != nil {
-			n.logger.Log("level", "error", "message", "failed to collect namespace consistency", "stack", fmt.Sprintf("%#v", err))
+		if helmclient.IsReleaseNotFound(err) {
+			continue
```
tomahawk28, line 78: If the release is not found, keep continuing to other releases.

service/collector/namespace_mismatch.go
```diff
@@ -70,11 +70,14 @@ func (n *NamespaceMismatch) Collect(ch chan<- prometheus.Metric) error {
...
+		} else if err != nil {
+			n.logger.Log("level", "warn", "message", "failed to collect namespace consistency", "stack", fmt.Sprintf("%#v", err))
```
tomahawk28, line 80: Change it to `warn` since it belongs to collector business.

rossf7 (approved):
LGTM

Should we have a fixed changelog?

tomahawk28:
> Should we have a fixed changelog?

Good call, I'm adding it now.

PR #602, opened 2020-09-24
Use checksum annotation and increase wait timeout

rossf7:
## Checklist


helm/chart-operator/values.yaml
```diff
@@ -10,7 +10,7 @@ helm:
   http:
     clientTimeout: "5s"
   kubernetes:
-    waitTimeout: "10s"
+    waitTimeout: "120s"
```
rossf7, line 13: 10s is very short for charts with hooks. So I think its better to use the same default we have in the catalog. WDYT?

tomahawk28 (approved):
LGTM, 👍

helm/chart-operator/values.yaml
tomahawk28: Sure, Let's go with 120s.

rossf7:
Tested on `godsmack`.

PR #616, opened 2020-10-05
Only wait for push to Aliyun for AWS

rossf7:
Sometimes the push to Aliyun can be slow and delays the other steps. We only need to wait when pushing to the AWS app collection.


tomahawk28 (approved):
Much better! Thanks

PR #618, opened 2020-10-12
Refactor addAnnotation

tomahawk28:
We are using patch annotation in quite a few places. It makes sense to refactor it and use it in general. 

Skipping update changelog since it's refactoring.

### DIFF
--- a/service/controller/chart/resource/release/create.go
+++ b/service/controller/chart/resource/release/create.go
@@ -100,7 +100,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		// The install will continue in the background. We set the checksum
 		// annotation so the update state calculation is accurate when we check
 		// in the next reconciliation loop.
-		err = r.patchAnnotations(ctx, cr, releaseState)
+		err = r.addHashAnnotation(ctx, cr, releaseState)
 		if err != nil {
 			return microerror.Mask(err)
 		}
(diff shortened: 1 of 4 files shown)


service/controller/chart/resource/release/resource.go
```diff
@@ -135,10 +135,66 @@ func (r *Resource) findHelmV2ConfigMaps(ctx context.Context, releaseName string)
 	return len(cms.Items) > 0, nil
 }
 
-// patchAnnotations updates the chart CR annotations if they have changed.
+func (r *Resource) addAnnotation(ctx context.Context, cr *v1alpha1.Chart, key, value string) error {
```
tomahawk28, line 138: create `addAnnotation` function in resource struct so we can use it in multiple places.

service/controller/chart/resource/release/create.go
```diff
@@ -156,7 +156,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 
 	// We set the checksum annotation so the update state calculation
```
rossf7, line 157:
```suggestion
	// We set the hash annotation so the update state calculation
```

service/controller/chart/resource/release/create.go
```diff
@@ -100,7 +100,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		// The install will continue in the background. We set the checksum
 		// annotation so the update state calculation is accurate when we check
 		// in the next reconciliation loop.
```
rossf7, line 102:
```suggestion
		// We set the hash annotation so the update state calculation is accurate
		// when we check in the next reconciliation loop.
```

Can we update the comment to match the name? 

The part about the installing continuing in the background is no longer accurate for Helm 3.

rossf7 (approved):
LGTM. 

Good to tidy this up. Just a couple of suggestions on the comments.

service/controller/chart/resource/release/create.go
tomahawk28: You're right, this is only true when we have a tiller pod. Changing it now.

PR #619, opened 2020-10-12
Stop updating revision in status in case of helm upgrade failure

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/13677

When chart-operator update the chart CR's status with the new helm revision number, it actually updates resourceVersion on the spec and induces another reconciliation, which triggers infinite helm upgrade behaviors. 

This is a patch which using `addStatusToContext` to not to update releaseRevision on the status update in case of helm failure. 

## Checklist

### DIFF
--- a/service/controller/chart/resource/status/create.go
+++ b/service/controller/chart/resource/status/create.go
@@ -33,8 +33,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		status := v1alpha1.ChartStatus{
 			Reason: cc.Status.Reason,
 			Release: v1alpha1.ChartStatusRelease{
-				LastDeployed: metav1.Time{},
-				Status:       cc.Status.Release.Status,
+				Status: cc.Status.Release.Status,
 			},
 		}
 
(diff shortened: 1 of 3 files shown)


tomahawk28:
Tested in `ginger`
Now it will show in status as below. 
```
status:
  appVersion: ""
  reason: 'Upgrade "aws-operator-9.1.2-dev-calvix" failed: failed to create resource:
    admission webhook "validation.gatekeeper.sh" denied the request: [denied by unique-operator-version]
    label app.kubernetes.io/version=9.1.2-dev-calvix is not unique in resources of
    kind "Deployment"'
  release:
    lastDeployed: null
    revision: 0
    status: failed
  version: ""
```
I don't know why `lastDeployed` is still showing `null`, but this is the quickest way to stop the infinite helm upgrade loop, WDYT?

service/controller/chart/resource/status/create.go
```diff
@@ -63,13 +64,6 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		if key.IsCordoned(cr) {
 			status = releaseStatusCordoned
 			reason = key.CordonReason(cr)
-		} else if cc.Status.Reason != "" {
```
tomahawk28, line 66: This statement is not necessary any longer.

rossf7 (approved):
LGTM

Damn that's a pain that adding the resource version caused this.

Good catch and the fix looks good.

> I don't know why lastDeployed is still showing null

I think it's because the CR status is set before GetReleaseContent is called. But the CR status is still set to failed so alerts will fire.

tomahawk28:
tested in `ginger`

PR #627, opened 2020-10-29
Update apiextensions to v3

rossf7:
## Checklist


service/controller/chart/resource/status/create.go
```diff
@@ -81,12 +82,14 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
-			Revision:     releaseContent.Revision,
-			Status:       status,
+			Revision: to.IntP(releaseContent.Revision),
+			Status:   status,
 		},
 		Version: releaseContent.Version,
 	}
+	if !releaseContent.LastDeployed.IsZero() {
+		desiredStatus.Release.LastDeployed = &metav1.Time{Time: releaseContent.LastDeployed}
+	}
```
rossf7, line 92: LastDeployed became an optional field in the CRD. So we need to handle nils or the operator crashes.

service/controller/chart/resource/status/create.go
```diff
@@ -130,11 +133,13 @@ func (r *Resource) setStatus(ctx context.Context, cr v1alpha1.Chart, status v1al
...
-		Reason:       status.Reason,
-		Status:       status.Release.Status,
-		Version:      status.Version,
+		AppVersion: status.AppVersion,
+		Reason:     status.Reason,
+		Status:     status.Release.Status,
+		Version:    status.Version,
+	}
+	if status.Release.LastDeployed != nil {
+		request.LastDeployed = *status.Release.LastDeployed
```
rossf7, line 142: Also for the webhook request.

rossf7:
Tested on `gauss`.

service/controller/chart/resource/status/create.go
rossf7: Ran into a problem in giraffe which is the CRD wasn't up to date and these optional fields were still required.

We should re-run `opsctl ensure crds` for the chart CRD before releasing this.

tomahawk28: Isn't that `chartcrd` resource in app-operator doing that? or is that resource only for TCs?

tomahawk28 (approved):
LGTM

PR #630, opened 2020-11-04
Using auth token for webhook

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/13201 

## Checklist

### DIFF
--- a/service/controller/chart/resource/status/create.go
+++ b/service/controller/chart/resource/status/create.go
@@ -104,8 +104,8 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 func (r *Resource) getAuthToken(ctx context.Context) (string, error) {
 	secret, err := r.k8sClient.CoreV1().Secrets(namespace).Get(ctx, authTokenName, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		// it is certainly control plane apps or auth secret is not created yet.
-		r.logger.LogCtx(ctx, "level", "debug", "message", "no auth token secret yet in cluster")
+		// There is no auth token secret. It may not have been created yet. Or the app CR is using InCluster.
+		r.logger.LogCtx(ctx, "level", "debug", "message", "no auth token secret found skip calling webhook")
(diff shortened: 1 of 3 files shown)


service/controller/chart/resource/status/create.go
```diff
@@ -100,9 +101,27 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	return nil
 }
 
+func (r *Resource) getAuthToken(ctx context.Context) (string, error) {
+	secret, err := r.k8sClient.CoreV1().Secrets(namespace).Get(ctx, authTokenName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		// it is certainly control plane apps or auth secret is not created yet.
+		r.logger.LogCtx(ctx, "level", "debug", "message", "no auth token secret yet in cluster")
```
tomahawk28, line 108: Logging if we can't find a secret would be nice, just to save our mental later time.

tomahawk28:
Tested in `gauss`

service/controller/chart/resource/status/create.go
rossf7, line 108:
```suggestion
		// There is no auth token secret. It may not have been created yet.
		r.logger.LogCtx(ctx, "level", "debug", "message", "no auth token secret found skip calling webhook")
```

Good to log this but can you also check if the app CR uses InCluster? If so we can skip this API call.

tomahawk28: Do you mean `chart` CRs? In chart-operator, I'm not sure how to check `InCluster` for chart CRs.

rossf7: Thanks you're right I was mixing up the CRs. There is no good way to detect that in the chart CR so this is fine, I updated my suggestion for the comment. WDYT?

tomahawk28: Perfect, thanks for your suggestion, I'm going to change that.

rossf7 (approved):
LGTM

PR #637, opened 2020-11-10
Adding app metadata

tomahawk28:
For testing https://github.com/giantswarm/giantswarm/issues/13973

## Checklist

### DIFF
--- a/helm/chart-operator/Chart.yaml
+++ b/helm/chart-operator/Chart.yaml
@@ -1,6 +1,11 @@
 apiVersion: v1
 name: chart-operator
-description: A Helm chart for the chart-operator
-home: https://github.com/giantswarm/chart-operator
-appVersion: [[ .AppVersion ]]
 version: [[ .Version ]]
+appVersion: [[ .AppVersion ]]
+home: https://github.com/giantswarm/chart-operator
+description: A Helm chart for the chart-operator
+restrictions:
+  clusterSingleton: true
+  namespaceSingleton: true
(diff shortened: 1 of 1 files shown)


helm/chart-operator/Chart.yaml
```diff
@@ -1,6 +1,11 @@
 apiVersion: v1
 name: chart-operator
-description: A Helm chart for the chart-operator
-home: https://github.com/giantswarm/chart-operator
-appVersion: [[ .AppVersion ]]
 version: [[ .Version ]]
+appVersion: [[ .AppVersion ]]
+home: https://github.com/giantswarm/chart-operator
```
rossf7, line 5: What about adding icon with the GS logo as well?

rossf7 (commented):
Looks good. Just wondering if we can add our GS icon. 

@cokiengchiara @oponder Do you have a link for that?

helm/chart-operator/Chart.yaml
oponder: not super important for these control plane apps, also we'll have a cool auto generated icon based on the first letters of the app name coming up soon for apps that don't define an icon.

unless you wanna make something clever or cute that builds on the giant swarm logo, for identifying it and building an identity for chart operator... for now I would leave it without an icon.

rossf7: Thanks @oponder cool lets leave it then.

tomahawk28:
going with a ping

PR #646, opened 2020-11-30
Fix comparison of last deployed and revision fields in chart CR status

rossf7:

We recently made last deployed and revision optional fields in the chart CR status but I screwed up the comparison in the status resource when it got integrated.

This caused chart-operator to OOM on `exodus`. As the status resource was updating the CR status on each loop and having problems calling the status webhook.

## Checklist

### DIFF
--- a/service/controller/chart/resource/status/create.go
+++ b/service/controller/chart/resource/status/create.go
@@ -12,6 +12,7 @@ import (
 	"github.com/giantswarm/helmclient/v3/pkg/helmclient"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/to"
+	"github.com/google/go-cmp/cmp"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
--- a/service/controller/chart/resource/status/create.go
+++ b/service/controller/chart/resource/status/create.go
@@ -12,7 +12,6 @@ import (
 	"github.com/giantswarm/helmclient/v3/pkg/helmclient"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/to"
-	"github.com/google/go-cmp/cmp"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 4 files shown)


service/controller/chart/resource/status/create.go
```diff
@@ -89,7 +89,10 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		Version: releaseContent.Version,
 	}
 	if !releaseContent.LastDeployed.IsZero() {
-		desiredStatus.Release.LastDeployed = &metav1.Time{Time: releaseContent.LastDeployed}
+		// We convert the timestamp to the nearest second to match the value in
+		// the chart CR status.
+		lastDeployed := releaseContent.LastDeployed.Unix()
+		desiredStatus.Release.LastDeployed = &metav1.Time{Time: time.Unix(lastDeployed, 0)}
```
rossf7, line 95: The last deployed timestamp we get from helm is to the nearest nanosecond. But the k8s API returns it to the nearest second.

This is just for info so the nearest second is fine.

service/controller/chart/resource/status/resource.go
```diff
@@ -83,13 +84,13 @@ func equals(a, b v1alpha1.ChartStatus) bool {
 	if a.AppVersion != b.AppVersion {
 		return false
 	}
-	if a.Release.LastDeployed != b.Release.LastDeployed {
+	if !a.Release.LastDeployed.Equal(b.Release.LastDeployed) {
```
rossf7, line 87: Using equal for comparison is safer.

service/controller/chart/resource/status/resource.go
```diff
@@ -83,13 +84,13 @@ func equals(a, b v1alpha1.ChartStatus) bool {
...
 		return false
 	}
 	if a.Reason != b.Reason {
 		return false
 	}
-	if a.Release.Revision != b.Release.Revision {
+	if to.Int(a.Release.Revision) != to.Int(b.Release.Revision) {
```
rossf7, line 93: This was all kinds of wrong. 🤦‍♂️ 

The revision became a pointer because it's optional. So we need to cast to an int for comparison.

rossf7:
Tested on `ginger`.

service/controller/chart/resource/status/create.go
tomahawk28: What do you think about adding a new unittest for comparing nanoseconds?

tomahawk28 (approved):
LGTM

service/controller/chart/resource/status/create.go
rossf7: Yes good idea. I updated `equals` to compare using the unix epoch which is to the nearest second. It just needed some handling in case its nil.

tomahawk28: Looks good to me :)

PR #650, opened 2020-12-03
Reduce requests and set memory limits

rossf7:

Reduces the requests as they are too high for tenant clusters. This was done previously by @corest in https://github.com/giantswarm/chart-operator/pull/543.

It also re-introduces the memory limit. If chart-operator is blocked by the API server it can create a lot of secrets which increases the memory usage. Without the memory limit the node can OOM.

The requests are too low for control planes. I'm testing out using VPA in control planes. If that doesn't work we can increase the requests via config.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,49 +63,6 @@ workflows:
             tags:
               only: /^v.*/
 
-      - architect/push-to-app-collection:
-          context: architect
-          name: push-chart-operator-to-aws-app-collection
-          app_name: "chart-operator"
-          app_collection_repo: "aws-app-collection"
-          unique: "true"
-          requires:
-            - push-chart-operator-to-aliyun
-            - push-chart-operator-to-control-plane-app-catalog
-          filters:
-            branches:
-              ignore: /.*/
-            tags:
-              only: /^v.*/
-
-      - architect/push-to-app-collection:
-          context: architect
-          name: push-chart-operator-to-azure-app-collection
-          app_name: "chart-operator"
-          app_collection_repo: "azure-app-collection"
-          unique: "true"
-          requires:
-            - push-chart-operator-to-control-plane-app-catalog
-          filters:
-            branches:
-              ignore: /.*/
-            tags:
-              only: /^v.*/
-
-      - architect/push-to-app-collection:
-          context: architect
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -63,49 +63,6 @@ workflows:
...
-          app_collection_repo: "kvm-app-collection"
-          unique: "true"
-          requires:
-            - push-chart-operator-to-control-plane-app-catalog
-          filters:
-            branches:
-              ignore: /.*/
-            tags:
-              only: /^v.*/
-
```
rossf7, line 108: Disabling the push to app collections due to customer change freeze. Will re-enable on Monday for the next release.

helm/chart-operator/values.yaml
```diff
@@ -1,11 +1,20 @@
 clusterDNSIP: 172.31.0.10
 
 e2e: false
-cnr:
-  address: https://quay.io
```
rossf7, line 5: Not needed anymore.

rossf7:
Tested on `ginger`.

helm/chart-operator/values.yaml
```diff
@@ -1,11 +1,20 @@
...
+cluster:
+  kubernetes:
+    domain: cluster.local
+
+deployment:
+  requests:
+    cpu: 50m
+    memory: 100Mi
+  limits:
+    memory: 250Mi
```
rossf7, line 16: This is the previous limit that was removed in https://github.com/giantswarm/chart-operator/pull/592

chart-operator OOMed in a couple of control planes but its likely this was due to the too many secrets problem.

.circleci/config.yml
tomahawk28: What do you think we introduce a button to trigger releasing to the app-collection? It's more flexible than what we used to do.

tomahawk28 (approved):

.circleci/config.yml
rossf7: That's a cool idea. We could add a hold step like for the e2e tests in azure-operator.

But usually we just want to release to the CPs and I'm going to add this back on Monday. We can do this if we need to disable the push to collections more often.

PR #653, opened 2020-12-04
Using mergo instead of helmclient for merging values

tomahawk28:
## Checklist

I don't think we need a changelog for this, all functions work as same.

### DIFF
--- a/integration/test/chart/basic/basic_test.go
+++ b/integration/test/chart/basic/basic_test.go
@@ -4,6 +4,7 @@ package basic
 
 import (
 	"context"
+	"reflect"
 	"testing"
 	"time"
 
@@ -12,9 +13,49 @@ import (
 	"github.com/giantswarm/backoff"
 	"github.com/giantswarm/helmclient/v3/pkg/helmclient"
 	"github.com/giantswarm/microerror"
+	"github.com/google/go-cmp/cmp"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
 
 	"github.com/giantswarm/chart-operator/v2/integration/key"
+)
+
+const (
+	configMapName = "test-app-configmap"
+	secretName    = "test-app-configmap"
+
+	configmapValue = `
+v1:
+  username: admin
+  host:
+    url: quay.io
+  memory_in_gb: 0
+  threshold: 2.17
+replicas: 3
+`
+	secretValue = `
+v1:
+  username: admin
+  host: 
+    secret: 
+      authToken: xer32wnq
+  githubToken: nnbhwk1dk
+  memory_in_gb: 3.14
+`
+
+	mergedValue = `
+v1:
+  username: admin
+  host:
+    url: quay.io
+    secret: 
+      authToken: xer32wnq
+  githubToken: nnbhwk1dk
+  memory_in_gb: 3.14
+  threshold: 2.17
+replicas: 3
+`
 )
 
 // TestChartLifecycle tests a Helm release can be created, updated and deleted
@@ -32,6 +73,47 @@ import (
 func TestChartLifecycle(t *testing.T) {
 	ctx := context.Background()
 
+	// creating dependant configmap & secret
+	{
+		config.Logger.Debugf(ctx, "creating configmap %#q", configMapName)
+
+		cr := &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      configMapName,
+				Namespace: key.Namespace(),
+			},
+			Data: map[string]string{
+				"values": configmapValue,
+			},
+		}
+
+		_, err := config.K8sClients.K8sClient().CoreV1().ConfigMaps(key.Namespace()).Create(ctx, cr, metav1.CreateOptions{})
+		if err != nil {
+			t.Fatalf("expected %#v got %#v", nil, err)
+		}
+
+		config.Logger.Debugf(ctx, "creating configmap %#q", configMapName)
+
+		config.Logger.Debugf(ctx, "creating secret %#q", secretName)
+
+		secret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      secretName,
+				Namespace: key.Namespace(),
+			},
+			StringData: map[string]string{
+				"values": secretValue,
+			},
+		}
+
+		_, err = config.K8sClients.K8sClient().CoreV1().Secrets(key.Namespace()).Create(ctx, secret, metav1.CreateOptions{})
+		if err != nil {
+			t.Fatalf("expected %#v got %#v", nil, err)
+		}
+
+		config.Logger.Debugf(ctx, "creating secret %#q", secretName)
+	}
+
 	// Test creation.
 	{
 		config.Logger.Debugf(ctx, "creating chart %#q", key.TestAppReleaseName())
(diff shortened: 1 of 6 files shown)


service/controller/chart/resource/release/resource.go
```diff
@@ -322,3 +322,19 @@ func toReleaseState(v interface{}) (ReleaseState, error) {
 
 	return *releaseState, nil
 }
+
+func convertFloat(m map[string]interface{}) {
+	for k, val := range m {
+		switch val.(type) {
+		case float64:
+			converted := int(val.(float64))
+			if val.(float64) == float64(converted) {
```
tomahawk28, line 331: If the decimal is the same without floating points, that means it should be an integer type.

service/controller/chart/resource/release/desired_test.go
```diff
@@ -289,16 +289,18 @@ func Test_DesiredState(t *testing.T) {
...
 			},
 			expectedState: ReleaseState{
 				Name:              "chart-operator-chart",
 				Status:            helmclient.StatusDeployed,
-				ValuesMD5Checksum: "40f61add2413651115ef94f85d8335cf",
+				ValuesMD5Checksum: "3b8440387b1462ecdceb25c4cb9ff065",
 				Values: map[string]interface{}{
 					"replicas":     2,
 					"secretnumber": 2,
+					"floatnumber":  3.14,
```
tomahawk28, line 303: Just to be sure it doesn't touch any float number.

tomahawk28:
Tested in `ginger`, no helm release upgrades are rolled out after this.

service/controller/chart/resource/release/desired.go
```diff
@@ -31,18 +33,20 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 	}
 
 	// Merge configmap and secret to provide a single set of values to Helm.
-	values, err := helmclient.MergeValues(configMapData, secretData)
+	err = mergo.Merge(&configMapData, secretData, mergo.WithOverride)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
 
+	convertFloat(configMapData)
```
rossf7, line 41: I think this needs a comment to explain why we need to convert the data.

Also can we be sure that converting the floats to integers won't cause problems in the components using these values?

rossf7 (commented):
I'm fine with switching to mergo but I'd like to see more test coverage.

So when there is a new mergo release we can be sure the merge logic is still working. What about adding an integration test for this?

service/controller/chart/resource/release/desired.go
tomahawk28: Yes, I checked merged values from control-plane apps, and floating values such as `memory_size_gb` are the same as before. 
```
 325               i3.xlarge:
 326                 cpu_cores: 4
 327                 description: I3 Storage Optimized Extra Large
 328                 memory_size_gb: 30.5
 329                 storage_size_gb: 950
 330               m3.2xlarge:
 331                 cpu_cores: 8
 332                 description: M3 General Purpose Double Extra Large
 333                 memory_size_gb: 30
 334                 storage_size_gb: 80
 335               m3.large:
 336                 cpu_cores: 2
 337                 description: M3 General Purpose Large
 338                 memory_size_gb: 7.5
 339                 storage_size_gb: 32
```
Other than that, integer values are not changed at all.

tomahawk28:
> What about adding an integration test for this? 

Good idea, but that means I need to pull values from helm release, right? What do you think about adding an additional function in helmclient to do that? perhaps we can get rid of those md5 annotation by using that function.

rossf7:
> Good idea, but that means I need to pull values from helm release, right? What do you think about adding an additional function in helmclient to do that? perhaps we can get rid of those md5 annotation by using that function.

I was thinking create a chart CR with a CM and secret with a complex merge. Then check the values for the helm release with `GetReleaseContent `.

The MD5 annotation was needed because the values we got back from `helm get values` was different to the values we provided. 

I'd prefer to keep it for now and just switch to mergo. Otherwise it's changing a lot of stuff at once.

tomahawk28:
> Then check the values for the helm release with GetReleaseContent .

Okay, I didn't know this magic before, let me try it.

service/controller/chart/resource/release/desired.go
rossf7: Ah OK  I think we get the EC2 instance data from an external API somehow.

Thanks, good news its just that!

integration/test/chart/basic/basic_test.go
```diff
@@ -32,6 +73,47 @@ import (
...
+				"values": configmapValue,
+			},
+		}
+
+		_, err := config.K8sClients.K8sClient().CoreV1().ConfigMaps(key.Namespace()).Create(ctx, cr, metav1.CreateOptions{})
+		if err != nil {
+			t.Fatalf("expected %#v got %#v", nil, err)
+		}
+
+		config.Logger.Debugf(ctx, "creating configmap %#q", configMapName)
```
rossf7, line 95:
```suggestion
		config.Logger.Debugf(ctx, "created configmap %#q", configMapName)
```

integration/test/chart/basic/basic_test.go
```diff
@@ -32,6 +73,47 @@ import (
...
+				"values": secretValue,
+			},
+		}
+
+		_, err = config.K8sClients.K8sClient().CoreV1().Secrets(key.Namespace()).Create(ctx, secret, metav1.CreateOptions{})
+		if err != nil {
+			t.Fatalf("expected %#v got %#v", nil, err)
+		}
+
+		config.Logger.Debugf(ctx, "creating secret %#q", secretName)
```
rossf7, line 114:
```suggestion
		config.Logger.Debugf(ctx, "created secret %#q", secretName)
```

service/controller/chart/resource/release/desired.go
```diff
@@ -38,6 +38,7 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
+	// Converting all float numbers to integer if it has the same value to return the same md5 hash.
```
rossf7, line 41:
```suggestion
	// Convert all floats to integers if they have the same value to return the same md5 hash.
```

integration/test/chart/basic/basic_test.go
```diff
@@ -12,11 +13,51 @@ import (
...
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
 
 	"github.com/giantswarm/chart-operator/v2/integration/key"
 )
 
+const (
+	configMapName = "test-app-configmap"
+	secretName    = "test-app-configmap"
```
rossf7, line 26:
```suggestion
	secretName    = "test-app-secret"
```

integration/test/chart/basic/basic_test.go
```diff
@@ -32,6 +73,47 @@ import (
 func TestChartLifecycle(t *testing.T) {
 	ctx := context.Background()
 
+	// creating dependant configmap & secret
```
rossf7, line 76:
```suggestion
	// Create dependant configmap & secret.
```

integration/test/chart/basic/basic_test.go
```diff
@@ -12,11 +13,51 @@ import (
...
+v1:
+  username: admin
+  host: 
+    secret: 
+      authToken: xer32wnq
+  githubToken: nnbhwk1dk
+  memory_in_gb: 3.14
+`
+
+	mergedValue = `
```
tomahawk28, line 47: I made values as complex as I could. @rossf7 PTAL

rossf7 (approved):
Nice, LGTM

Just a few minor things.

integration/test/chart/basic/basic_test.go
tomahawk28: Damn :(  I need more coffee today morning.

integration/test/chart/basic/basic_test.go
rossf7: These look good. Most types are covered. What about adding a boolean? Maybe also some quoted strings? 

Basically anything in YAML that is a bit weird. Which is most things. 😄

tomahawk28: Good call, I would add both cases.

tomahawk28:
Thanks for your input! I'm going to merge it now.

PR #658, opened 2020-12-07
Add VPA flag to values and re-enable push to app collections

rossf7:

Fixes the VPA templating so it can be controlled via the chart values. This is so we can disable VPA in integration tests.

Also re-enables pushing to app collections which was disabled due to a customer change freeze.

## Checklist


rossf7:

PR #670, opened 2020-12-22
Adding diff logs into release resource

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/14626

Printing release differences make our lives easier (to debug it)

## Checklist

### DIFF
--- a/service/controller/chart/resource/release/update.go
+++ b/service/controller/chart/resource/release/update.go
@@ -249,7 +249,7 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
 		}, cmp.Ignore())
 
 		if diff := cmp.Diff(currentReleaseState, desiredReleaseState, opt); diff != "" {
-			fmt.Printf("release %#q have to be updated, (-current +desired):\n%s", cr.Name, diff)
+			fmt.Printf("release %#q has to be updated, (-current +desired):\n%s", cr.Name, diff)
 		}
 
 		return &desiredReleaseState, nil
--- a/service/controller/chart/resource/release/update.go
+++ b/service/controller/chart/resource/release/update.go
@@ -244,7 +244,14 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
 	}
 
 	if isReleaseModified(currentReleaseState, desiredReleaseState) {
-		if diff := cmp.Diff(currentReleaseState, desiredReleaseState); diff != "" {
+		opt := cmp.FilterPath(func(p cmp.Path) bool {
+			if p.String() == "Values" {
+				return true
+			}
+			return false
+		}, cmp.Ignore())
+
+		if diff := cmp.Diff(currentReleaseState, desiredReleaseState, opt); diff != "" {
 			fmt.Printf("release %#q have to be updated, (-current +desired):\n%s", cr.Name, diff)
 		}
 
--- a/service/controller/chart/resource/release/update.go
+++ b/service/controller/chart/resource/release/update.go
@@ -10,6 +10,7 @@ import (
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/operatorkit/v4/pkg/controller/context/resourcecanceledcontext"
 	"github.com/giantswarm/operatorkit/v4/pkg/resource/crud"
+	"github.com/google/go-cmp/cmp"
 
 	"github.com/giantswarm/chart-operator/v2/pkg/annotation"
 	"github.com/giantswarm/chart-operator/v2/service/controller/chart/controllercontext"
@@ -243,10 +244,11 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
 	}
 
 	if isReleaseModified(currentReleaseState, desiredReleaseState) {
-		r.logger.Debugf(ctx, "the %#q release has to be updated", desiredReleaseState.Name)
+		if diff := cmp.Diff(currentReleaseState, desiredReleaseState); diff != "" {
+			fmt.Printf("release %#q have to be updated, (-current +desired):\n%s", cr.Name, diff)
+		}
+
 		return &desiredReleaseState, nil
-	} else {
-		r.logger.Debugf(ctx, "the %#q release does not have to be updated", desiredReleaseState.Name)
 	}
 
 	err = r.removeAnnotation(ctx, &cr, annotation.RollbackCount)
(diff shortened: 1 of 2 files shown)


tomahawk28:
I didn't expect it would print out values difference as below
```
release `vertical-pod-autoscaler-app-unique` have to be updated, (-current +desired):
  release.ReleaseState{
        Name:              "vertical-pod-autoscaler-app-unique",
        Status:            "deployed",
-       ValuesMD5Checksum: "",
+       ValuesMD5Checksum: "052155a75f19932165fb73a8f16c02f0",
-       Values:            nil,
+       Values: map[string]interface{}{
+               "Installation": map[string]interface{}{
+                       "V1": map[string]interface{}{
+                               "Auth":       map[string]interface{}{...},
+                               "Customer":   string("giantswarm"),
+                               "GiantSwarm": map[string]interface{}{...},
+                               "Guest":      map[string]interface{}{...},
+                               ...
+                       },
+               },
+               "admissionController":   map[string]interface{}{"enabled": bool(true), "replicaCount": int(2)},
+               "alertmanager":          map[string]interface{}{"enabled": bool(false)},
+               "coreDns":               map[string]interface{}{"enabled": bool(false)},
+               "defaultRules":          map[string]interface{}{"create": bool(false)},
+               "grafana":               map[string]interface{}{"enabled": bool(false)},
+               "kubeApiServer":         map[string]interface{}{"enabled": bool(false)},
+               "kubeControllerManager": map[string]interface{}{"enabled": bool(false)},
+               "kubeEtcd":              map[string]interface{}{"enabled": bool(false)},
+               "kubeProxy":             map[string]interface{}{"enabled": bool(false)},
+               "kubeScheduler":         map[string]interface{}{"enabled": bool(false)},
+               "kubeStateMetrics":      map[string]interface{}{"enabled": bool(false)},
+               "kubelet":               map[string]interface{}{"enabled": bool(false)},
+               "nodeExporter":          map[string]interface{}{"enabled": bool(false)},
+               "prometheus":            map[string]interface{}{"enabled": bool(false)},
+               "prometheusOperator": map[string]interface{}{
+                       "configReloaderCpu":    string("25m"),
+                       "createCustomResource": bool(false),
+                       "kubeletService":       map[string]interface{}{"enabled": bool(false)},
+                       "resources":            map[string]interface{}{"limits": map[string]interface{}{...}, "requests": map[string]interface{}{...}},
+                       "serviceMonitor":       map[string]interface{}{"selfMonitor": bool(false)},
+               },
+               ...
+       },
-       Version: "1.0.1-19aa2842a2bec33de831503fa0386a16504b3c1f",
+       Version: "1.0.3",
  }
```
But since we might include secret data, I don't think it's a good idea.

tomahawk28:
Ignoring `Values` for security purpose. 
```
D 12/22 12:00:28 /apis/application.giantswarm.io/v1alpha1/namespaces/giantswarm/charts/dex-app-unique release.NewUpdatePatch finding out if the `dex-app-unique` release has to be updated | micrologger@v0.4.0/logger.go:53 | controller=chart-operator-chart | event=update | loop=69 | version=176156657
release `dex-app-unique` have to be updated, (-current +desired):
  release.ReleaseState{
        Name:              "dex-app-unique",
        Status:            "deployed",
-       ValuesMD5Checksum: "",
+       ValuesMD5Checksum: "1678b4446ba0392da6681840add3d06a",
        ... // 1 ignored field
-       Version: "1.5.0",
+       Version: "1.2.2-7fbf6691d0804220fb4802c6042c3c9750e1c2a0",
  }
```

service/controller/chart/resource/release/update.go
```diff
@@ -243,10 +244,15 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
 	}
 
 	if isReleaseModified(currentReleaseState, desiredReleaseState) {
-		r.logger.Debugf(ctx, "the %#q release has to be updated", desiredReleaseState.Name)
+		opt := cmp.FilterPath(func(p cmp.Path) bool {
+			return p.String() == "Values"
```
tomahawk28, line 248: This is to ignore `Values` at all, I tried to rename it to `values` to unable to exported but it brought broader changes that I didn't mean to.

tomahawk28:
Tested in `ginger`

service/controller/chart/resource/release/update.go
```diff
@@ -243,10 +244,15 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
...
+		}, cmp.Ignore())
+
+		if diff := cmp.Diff(currentReleaseState, desiredReleaseState, opt); diff != "" {
+			fmt.Printf("release %#q have to be updated, (-current +desired):\n%s", cr.Name, diff)
+		}
+
 		return &desiredReleaseState, nil
-	} else {
-		r.logger.Debugf(ctx, "the %#q release does not have to be updated", desiredReleaseState.Name)
```
tomahawk28, line 249: I don't think it has any value to print messages like above. I need to see the whole sentence to just see that it's not gonna do anything.

service/controller/chart/resource/release/update.go
rossf7: What happens if just the Values are changed?

Good to exclude the Values from the diff but we should log that somehow.

service/controller/chart/resource/release/update.go
rossf7: Good point it doesn't add much value. 👍

service/controller/chart/resource/release/update.go
tomahawk28: I think md5 would be different then, we would show md5 hash into the difference. I think that's enough information we could show, WDYT?

rossf7: Yes totally. I forgot the MD5 hash will appear in the diff. 😅 

What about adding a comment to explain that? As its easy to miss when reading quickly.

rossf7 (approved):
LGTM

service/controller/chart/resource/release/update.go
tomahawk28: Good call, I added a comment as you suggested.

service/controller/chart/resource/release/update.go
```diff
@@ -243,10 +244,16 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
 	}
 
 	if isReleaseModified(currentReleaseState, desiredReleaseState) {
-		r.logger.Debugf(ctx, "the %#q release has to be updated", desiredReleaseState.Name)
+		// Ignoring `Values` among diff since it could contain secret data.
```
rossf7, line 247:
```suggestion
		// Ignoring `Values` in diff since it could contain secret data and we use MD5 hash for comparison.
```

Could we go with this?

tomahawk28: Nice :)

PR #721, opened 2021-03-19
Enable CNI app installation

kubasobon:
Towards https://github.com/giantswarm/giantswarm/issues/16283

:construction_worker: This branch should **not** be merged. It serves as a base for CAPI release with modified `chart-operator`. 

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  architect: giantswarm/architect@2.3.0
+  architect: giantswarm/architect@2.9.0
 
 workflows:
   build:
@@ -110,7 +110,7 @@ workflows:
               ignore: /.*/
             tags:
               only: /^v.*/
-              
+
       - architect/push-to-app-collection:
           context: architect
           name: push-chart-operator-to-vmware-app-collection
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  architect: giantswarm/architect@2.9.0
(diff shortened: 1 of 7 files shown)


QuentinBisson (approved):

MarcelMue (approved):

anvddriesch:
What is the status here? Can it either be merged or tagged as we did we cluster-operator so we can use it in our CAPA test release?

rossf7:
> What is the status here? Can it either be merged or tagged as we did we cluster-operator so we can use it in our CAPA test release?

@anvddriesch @kubasobon Could we tag it for now?

We need to be sure this also works with existing clusters. I think we can do this by setting values in the cluster configmap for chart-operator.

anvddriesch:
You mean simply tag it with something like `x.x.x-capi` as we did with cluster-operator? That's completely fine from my side. Then we an only use it in CAPI releases and it doesn't have to work with other clusters yet.

rossf7:
@anvddriesch Yes, I think we could tag it as `2.11.0-capi` as the changes are based on that release.

This is fine for k8s 1.19. For 1.20 we need at least `2.13.1` as there was a problem with operatorkit.

@kubasobon WDYT?

anvddriesch:
k8s 1.19 is the version we currently use for the CAPA test release so that's fine. :D

kubasobon:
On it.

kubasobon:
https://github.com/giantswarm/chart-operator/releases/tag/v2.11.0-capi-1

rossf7:
@kubasobon Same this is also done. Changes were released in chart-operator 2.16.0.

PR #745, opened 2021-04-30
Handling unknown helm errors

tomahawk28:
When chart-operator handles unknown helm errors, it didn't update its status so the previous status has been preserved.   

Even though these changes would show users `unknown-error`, this is still far better than being unnoticed. 

## Checklist

### DIFF
--- a/service/controller/chart/resource/release/create.go
+++ b/service/controller/chart/resource/release/create.go
@@ -140,29 +140,35 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		resourcecanceledcontext.SetCanceled(ctx)
 		return nil
 	} else if err != nil {
-		reason := err.Error()
+		//reason := err.Error()
 		r.logger.Errorf(ctx, err, "helm release %#q failed", releaseState.Name)
 
-		releaseContent, err := r.helmClient.GetReleaseContent(ctx, ns, releaseState.Name)
-		if helmclient.IsReleaseNotFound(err) {
-			reason = fmt.Sprintf("helm error: (%s)", reason)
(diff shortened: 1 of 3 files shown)


service/controller/chart/resource/release/create.go
```diff
@@ -140,29 +140,34 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
...
 		// Release is failed so the status resource will check the Helm release.
 		if releaseContent.Status == helmclient.StatusFailed {
+			addStatusToContext(cc, releaseContent.Description, helmclient.StatusFailed)
+
 			r.logger.Debugf(ctx, "failed to create release %#q", releaseContent.Name)
 			r.logger.Debugf(ctx, "canceling resource")
 			resourcecanceledcontext.SetCanceled(ctx)
 			return nil
 		}
-		return microerror.Mask(err)
```
tomahawk28, line 165: Rather than returning errors, we should let users know about the exact what happening by updating the status resources.

service/controller/chart/resource/release/resource.go
```diff
@@ -48,6 +48,9 @@ const (
 	// not apply for failures.
 	releaseFailedMaxAttempts = 5
 
+	// unknownError when a release fails by unknown reasons.
```
rossf7, line 51:
```suggestion
	// unknownError when a release fails for unknown reasons.
```

service/controller/chart/resource/release/create.go
rossf7: Nice catch and yes much better to show this in the status.

rossf7 (approved):
LGTM

PR #758, opened 2021-05-28
Allow installing CNI apps for CAPI support

rossf7:
Towards https://github.com/giantswarm/roadmap/issues/320

This is so we can merge the changes from @kubasobon in https://github.com/giantswarm/chart-operator/pull/721

To enable these settings the cluster values configmap will have `chartOperator.cni.install` set to true. Otherwise we use the current settings.

## Checklist

### DIFF
--- a/helm/chart-operator/values.yaml
+++ b/helm/chart-operator/values.yaml
@@ -1,3 +1,8 @@
+# For CAPI clusters this will be set to true. So charts for CNI apps can be installed.
+chartOperator:
+  cni:
+    install: false
+
 cluster:
   kubernetes:
     domain: cluster.local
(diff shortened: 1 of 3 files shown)


helm/chart-operator/values.yaml
```diff
@@ -1,3 +1,8 @@
+# For CAPI clusters this will be set to true. So charts for CNI apps can be installed.
+chartOperator:
```
rossf7, line 2: The chartOperator here is a bit redundant. But it's so it's clear in the cluster values configmap that will be managed by cluster-apps-operator.

anvddriesch (approved):

tomahawk28 (approved):

kubasobon (approved):
:tada:

rossf7:
Tested on gauss.

PR #762, opened 2021-06-02
Create priority class if CNI install is true

rossf7:
Towards giantswarm/roadmap#320

## Checklist


helm/chart-operator/templates/priorityclass.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.e2e  }}
+{{- if or .Values.chartOperator.cni.install .Values.e2e  }}
```
rossf7, line 1: The priority class is created via ignition but that won't work for CAPI clusters.

rossf7:
Tested on ginger. Going with a ping.

PR #765, opened 2021-06-03
Migrate to configuration management

corest:
Towards https://github.com/giantswarm/giantswarm/issues/17386

## Checklist

### DIFF
--- a/helm/chart-operator/templates/deployment.yaml
+++ b/helm/chart-operator/templates/deployment.yaml
@@ -51,7 +51,7 @@ spec:
       - key: node-role.kubernetes.io/master
         effect: NoSchedule
       {{- end }}
-      {{ if (.Values.Installation) }}
+      {{ if .Values.isManagementCluster }}
       dnsPolicy: ClusterFirst
       {{ else }}
       dnsPolicy: None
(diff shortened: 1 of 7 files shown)


rossf7 (approved):
LGTM

helm/chart-operator/values.yaml
```diff
@@ -70,3 +73,5 @@ tiller:
 
 verticalPodAutoscaler:
   enabled: true
+
+isManagementCluster: false
```
kopiczko, line 77: Sneaky. We can do the same for all apps that rely on the `-unique` suffix.

helm/chart-operator/templates/deployment.yaml
```diff
@@ -70,7 +70,11 @@ spec:
           value: "5"
       initContainers:
       - name: wait-for-quay
+        {{ if .Values.isManagementCluster }}
+        image: "{{ .Values.registry.domain }}/{{ .Values.image.name }}:{{ .Values.image.tag }}"
+        {{ else }}
         image: "{{ .Values.image.registry }}/{{ .Values.image.name }}:{{ .Values.image.tag }}"
+        {{ end }}
```
kopiczko, line 77: For my interest: why different images on MC and WC?

kopiczko (approved):

helm/chart-operator/templates/deployment.yaml
corest: I don't know, but also don't want to investigate this now. Can be easily changed via config repo later

rossf7: We use `image.registry` in the default catalog. So I'd prefer to use that and remove some of the conditionals.

But I'm fine if we do that in batman once this migration is done.

https://github.com/giantswarm/installations/blob/80e965ba6c529a52f491932fb0972e851c29b374/giraffe/appcatalog/default-appcatalog-values.yaml#L10-L11

PR #766, opened 2021-06-04
Improve status message when helm release has failed max number of times

rossf7:
Fixes https://github.com/giantswarm/giantswarm/issues/17314

When chart-operator fails to update a helm release 5 times we stop updating. This is because the max history of 10 revisions is not respected and 100s of release secrets can be created which impacts the cluster.

This improves the CR status so its clearer when this has happened.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 - Prepare helm values to configuration management.
 - Update architect-orb to v3.0.0.
 
+### Fixed
+
+- Improve status message when helm release has failed max number of attempts.
+
 ## [2.16.0] - 2021-06-03
 
 ### Changed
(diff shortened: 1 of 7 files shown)


rossf7:
Tested on gauss.

```sh
kubectl -n giantswarm get app prometheus-operator-app-unique -o yaml  | yq .status.release
{
  "lastDeployed": "2021-06-07T06:55:26Z",
  "reason": "Release has failed 5 times and will not be updated.\nReason: Upgrade \"prometheus-operator-app-unique\" failed: pre-upgrade hooks failed: warning: Hook pre-upgrade prometheus-operator-app/templates/prometheus-operator/admission-webhooks/job-patch/rolebinding.yaml failed: object is being deleted: rolebindings.rbac.authorization.k8s.io \"prometheus-operator-app-un-admission\" already exists",
  "status": "failed"
}
```

tomahawk28 (approved):
LGTM!

PR #771, opened 2021-06-16
Retry upgrades even if values or version are unchanged and add releasemaxfailed resource

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/17622

## Checklist

### DIFF
--- a/service/controller/chart/resource/namespace/resource.go
+++ b/service/controller/chart/resource/namespace/resource.go
@@ -34,7 +34,7 @@ type Resource struct {
 	k8sWaitTimeout time.Duration
 }
 
-// New creates a new configured tcnamespace resource.
+// New creates a new configured namespace resource.
 func New(config Config) (*Resource, error) {
 	// Dependencies.
 	if config.K8sClient == nil {
--- a/service/controller/chart/resource/release/resource.go
+++ b/service/controller/chart/resource/release/resource.go
@@ -261,34 +261,6 @@ func isEmpty(c ReleaseState) bool {
 	return equals(c, ReleaseState{})
 }
 
-// isReleaseFailed checks if the release is failed. If the values or version
-// has changed we return false and will attempt to update the release. As this
-// may fix the problem.
-func isReleaseFailed(current, desired ReleaseState) bool {
-	result := false
-
-	if !isEmpty(current) {
-		// Values have changed so we should try to update even if the release
-		// is failed.
-		if current.ValuesMD5Checksum != desired.ValuesMD5Checksum {
-			return false
-		}
-
-		// Version has changed so we should try to update even if the release
-		// is failed.
-		if current.Version != desired.Version {
-			return false
-		}
-
-		// Release is failed and should not be updated.
-		if current.Status == helmclient.StatusFailed {
-			result = true
-		}
-	}
-
-	return result
-}
-
 func isReleaseInTransitionState(r ReleaseState) bool {
 	return helmclient.ReleaseTransitionStatuses[r.Status]
 }
--- a/service/controller/chart/resource/release/resource.go
+++ b/service/controller/chart/resource/release/resource.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"sort"
 	"strings"
 	"time"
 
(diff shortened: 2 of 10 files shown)


service/controller/chart/resource/namespace/resource.go
```diff
@@ -34,7 +34,7 @@ type Resource struct {
 	k8sWaitTimeout time.Duration
 }
 
-// New creates a new configured tcnamespace resource.
+// New creates a new configured namespace resource.
```
rossf7, line 37: Just fixing this.

service/controller/chart/resource/release/resource.go
```diff
@@ -230,34 +228,6 @@ func (r *Resource) addHashAnnotation(ctx context.Context, cr v1alpha1.Chart, rel
...
-
-	for i := 0; i < project.ReleaseFailedMaxAttempts; i++ {
-		if history[i].Status != helmclient.StatusFailed {
-			return false, nil
-		}
-	}
-
-	// All failed so we exceeded the max attempts.
-	return true, nil
-}
```
rossf7, line 259: This moves to the new releasemaxhistory resource.

service/controller/chart/resource/release/update.go
```diff
@@ -254,28 +254,17 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
...
-	if isReleaseFailed(currentReleaseState, desiredReleaseState) {
-		r.logger.Debugf(ctx, "the %#q release is in status %#q and cannot be updated", desiredReleaseState.Name, currentReleaseState.Status)
+	// We check the controller context and if the release has failed more
+	// than the max number of attempts we stop updating. Otherwise too many
+	// release secrets will be created.
+	//
+	// The releasemaxhistory resource handles deleting the oldest secret if
+	// it is over 1 minute old. So we still retry but at a slower rate.
+	if currentReleaseState.Status == helmclient.StatusFailed && cc.Status.Release.FailedMaxAttempts {
+		r.logger.Debugf(ctx, "the %#q release is in status %#q and has failed %d times", desiredReleaseState.Name, currentReleaseState.Status, project.ReleaseFailedMaxAttempts)
```
rossf7, line 264: We should retry even if values or version are unchanged for situations like failed webhooks.

This should mean we get less pages but we also slow the retries and still ensure that too many release secrets are not created.

service/controller/chart/resources.go
```diff
@@ -135,9 +151,15 @@ func newChartResources(config chartResourcesConfig) ([]resource.Interface, error
...
 	resources := []resource.Interface{
+		// tiller migration deletes tiller once helm 3 migration is done.
 		tillerMigrationResource,
+		// namespace creates the release namespace and allows setting metadata.
 		namespaceResource,
+		// release max history ensures not too many helm release secrets are created.
+		releaseMaxHistoryResource,
+		// release manages Helm releases and is the most important resource.
 		releaseResource,
+		// status resource manages the chart CR status.
```
rossf7, line 162: We have these comments in other operators and I think are useful here too. 🤞 we can delete tiller migration soon.

rossf7 (commented):
Tested on gorgoth

service/controller/chart/resource/releasemaxhistory/create.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+	if len(secrets.Items) != 1 {
+		return false, microerror.Maskf(executionFailedError, "expected 1 release secret got %d", len(secrets.Items))
+	}
+
+	secret := secrets.Items[0]
+	diff := time.Since(secret.CreationTimestamp.Time)
+	if diff.Minutes() < 1 {
```
rossf7, line 94: I think 1 minute is enough to ensure upgrades are rate limited. But maybe longer is needed. @tomahawk28 WDYT?

tomahawk28: Sure, 1 minute is reasonable to wait.

service/controller/chart/resource/releasemaxhistory/create.go
```diff
@@ -0,0 +1,143 @@
...
+		r.logger.Debugf(ctx, "revision %d for release %#q is < 1 minutes old", rev.Revision, releaseName)
+		r.logger.Debugf(ctx, "canceling resource")
```
tomahawk28, line 96: I think we are not canceling the resource at this moment, just returning `false`.

rossf7: Good catch, thx! No enough ☕ when I did this

rossf7:
I did some more testing with workload clusters and looks fine.

@tomahawk28 PTAL

tomahawk28 (approved):
LGTM!

PR #778, opened 2021-06-22
Parametrize `dig` with used image registry

kubasobon:
Towards https://github.com/giantswarm/giantswarm/issues/17459

This is especially important in China. Example:
```
chart-operator-unique-f7d49ff56-m5mhs   0/1     Init:0/1   0          27m
```

## Checklist

### DIFF
--- a/helm/chart-operator/templates/deployment.yaml
+++ b/helm/chart-operator/templates/deployment.yaml
@@ -70,19 +70,11 @@ spec:
           value: "5"
       initContainers:
       - name: wait-for-quay
-        {{ if .Values.isManagementCluster }}
-        image: "{{ .Values.registry.domain }}/{{ .Values.image.name }}:{{ .Values.image.tag }}"
-        {{ else }}
         image: "{{ .Values.image.registry }}/{{ .Values.image.name }}:{{ .Values.image.tag }}"
-        {{ end }}
         command:
         - sh
         - -c
-        {{ if .Values.isManagementCluster }}
-        - until dig {{ .Values.registry.domain }}. {{ range (splitList "," .Values.externalDNSIP) }}@{{ . }} {{ end }}; do echo waiting for {{ .Values.registry.domain }}; sleep 2; done;
-        {{ else }}
         - until dig {{ .Values.image.registry }}. {{ range (splitList "," .Values.externalDNSIP) }}@{{ . }} {{ end }}; do echo waiting for {{ .Values.image.registry }}; sleep 2; done;
-        {{ end }}
         securityContext:
           runAsUser: {{ .Values.pod.user.id }}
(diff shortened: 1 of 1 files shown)


helm/chart-operator/templates/deployment.yaml
```diff
@@ -78,7 +78,11 @@ spec:
         command:
         - sh
         - -c
-        - until dig quay.io. {{ range (splitList "," .Values.externalDNSIP) }}@{{ . }} {{ end }}; do echo waiting for quay.io; sleep 2; done;
+        {{ if .Values.isManagementCluster }}
+        - until dig {{ .Values.registry.domain }}. {{ range (splitList "," .Values.externalDNSIP) }}@{{ . }} {{ end }}; do echo waiting for {{ .Values.registry.domain }}; sleep 2; done;
```
kubasobon, line 82: `.Values.externalDNSIP` defaulted to `8.8.8.8` in China for me. I'm not sure this is right.

tomahawk28:
Your change seems fine but I found out we actually have no `initContainers` in the spec of chart-operator in the management cluster. I think we have an issue with the template in the first place because I can't see it from `helm manifest`.
@kubasobon Do you mind if I patch it in this branch?
https://github.com/giantswarm/chart-operator/blob/a3a2f123a9f929f5929ed17c12816515e6348f36/helm/chart-operator/templates/deployment.yaml#L71-L85

kubasobon:
@tomahawk28 I don't mind if it's patched elsewhere, but 2 questions:
1. Can we still change `dig quay.io` to region appropriate repository? No China cluster should depend on resolving `quay.io`, even workload clusters.
2.  What value/values am I missing in akita's configuration that I got `initContainers` in MC? Link: https://github.com/giantswarm/installations/pull/1816

tomahawk28:
> 1. Can we still change dig quay.io to region appropriate repository? 

Yes, we can change that. 

> 2. What value/values am I missing in akita's configuration that I got initContainers in MC? 

I talked with @rossf7 and we concluded that we don't need `initContainers` in MC actually. 
I think we are okay but you can delete the 3 lines below to make the code simpler. 
https://github.com/giantswarm/chart-operator/blob/a3a2f123a9f929f5929ed17c12816515e6348f36/helm/chart-operator/templates/deployment.yaml#L73-L75

kubasobon:
I removed `managementCluster` parametrization.

helm/chart-operator/templates/deployment.yaml
tomahawk28: I think it's better to ask @giantswarm/team-firecracker about this. Can we still use 8.8.8.8 in the China installations?

tomahawk28 (approved):
LGTM :)

PR #832, opened 2021-12-15
Use apptestctl to install CRDs due to GitHub rate limits

rossf7:
Same as https://github.com/giantswarm/app-operator/pull/892

## Checklist


integration/setup/setup.go
```diff
@@ -84,7 +84,6 @@ func installResources(ctx context.Context, config Config) error {
 		}
 		values := map[string]interface{}{
 			"clusterDNSIP": "10.96.0.10",
-			"e2e":          "true",
```
rossf7, line 87: Same deal with the priority class its now created by apptestctl

MarcelMue (approved):

PR #844, opened 2022-02-04
Change the default port to 8150

kopiczko:


rossf7 (approved):
Makes sense 👍

pipo02mix (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ### Changed
 
+- Change the default port to 8150 to avoid port conflicts in the CAPI setup.
```
MarcelMue, line 12: What is actually conflicting with this port?

I am not sure if there is something else (e.g. monitoring) depending on this port

tfussell: It should be scraped via the service https://github.com/giantswarm/chart-operator/blob/master/helm/chart-operator/templates/service.yaml#L10 so I don't think monitoring would be affected. Still might be worth confirming.

We use `hostNetwork` because it needs to run without CNI.

MarcelMue (approved):
Let's do it - I am tired of restarting chart-operator :cry:

rossf7:
> What do I still need to (besides a release) do to deploy it with the new openstack WCs?

You need to bump the version in the cluster-apps-operator helm chart and release a new version.

https://github.com/giantswarm/cluster-apps-operator/blob/0efb60a84f29e04802a8c4b4ecbf22a50cec83a1/helm/cluster-apps-operator/values.yaml#L7

PR #854, opened 2022-03-01
Add retry when getting chart-operator tarball URL

rossf7:


integration/setup/setup.go
```diff
@@ -47,17 +49,32 @@ func installResources(ctx context.Context, config Config) error {
...
+	var operatorTarballURL string
 	{
-		config.Logger.Debugf(ctx, "getting tarball URL")
+		config.Logger.Debugf(ctx, "getting %#q tarball URL", project.Name())
+
+		o := func() error {
+			operatorTarballURL, err = appcatalog.GetLatestChart(ctx, key.DefaultTestCatalogStorageURL(), project.Name(), env.CircleSHA())
+			if err != nil {
+				return microerror.Mask(err)
+			}
```
rossf7, line 60: Same as in app-operator we can't install chart-operator as an app. So we need a retry here.

MarcelMue (approved):
LET'S GOOO retries save the world

rossf7:
MOAR RETRIES 😄

PR #873, opened 2022-05-17
Correctly update App status on some deployment failures

uvegla:
Towards: https://github.com/giantswarm/giantswarm/issues/22017

## Checklist

### DIFF
--- a/service/controller/chart/resource/status/create.go
+++ b/service/controller/chart/resource/status/create.go
@@ -35,6 +35,8 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	releaseName := key.ReleaseName(cr)
 	r.logger.Debugf(ctx, "getting status for release %#q", releaseName)
 
+	r.logger.Debugf(ctx, "the controller context for %#q is: %#q", releaseName, cc)
+
 	releaseContent, err := r.helmClient.GetReleaseContent(ctx, key.Namespace(cr), releaseName)
 	if helmclient.IsReleaseNotFound(err) {
 		r.logger.Debugf(ctx, "release %#q not found", releaseName)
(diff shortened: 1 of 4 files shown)


service/controller/chart/resource/status/create.go
```diff
@@ -35,6 +35,8 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 	releaseName := key.ReleaseName(cr)
 	r.logger.Debugf(ctx, "getting status for release %#q", releaseName)
 
+	r.logger.Debugf(ctx, "the controller context for %#q is: %#q", releaseName, cc)
```
kubasobon, line 38: Is it useful when debugging the Chart CR?

service/controller/chart/resource/status/create.go
```diff
@@ -73,27 +75,32 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
 		}
 	}
 
 	desiredStatus := v1alpha1.ChartStatus{
-		AppVersion: releaseContent.AppVersion,
-		Reason:     reason,
-		Release: v1alpha1.ChartStatusRelease{
-			Revision: to.IntP(releaseContent.Revision),
-			Status:   status,
+		AppVersion: releaseContent.AppVersion, // The desired should depend on the version set in the Chart, not what is deployed by Helm
```
kubasobon, line 97: There are so many questions here in the comments. Do you want to merge it like that, or would you rather wait for them to get answered?

uvegla: Right, no it is a leftover if making sense of it all.

service/controller/chart/resource/status/create.go
uvegla: Nope, it will end up in the status reason now, cleaned it up.

MarcelMue (approved):

PR #875, opened 2022-05-19
Introducing split client approach for Helm releases

ljakimczuk:
## Checklist


flag/service/helm/helm.go
```diff
@@ -6,8 +6,15 @@ import (
...
-	MaxRollback     string
+	HTTP        http.HTTP
+	Kubernetes  kubernetes.Kubernetes
+	MaxRollback string
+
+	// SplitClient determines usage of additional pubHelmClient impersonating
+	// `default:automation` Service Account for App CRs created outside the
+	// `giantswarm` namespace. When `false` Chart Operator runs under full
+	// cluster admin permissions no matter the App CR namespace.
+	SplitClient string
```
kubasobon, line 17: It's defined as `bool` [here](https://github.com/giantswarm/chart-operator/pull/875/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R112). Which is it?

kubasobon (approved):

service/internal/clientpair/clientpair.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/giantswarm/micrologger"
+
+	"github.com/giantswarm/chart-operator/v2/service/controller/chart/key"
+)
+
+const (
+	// privateNamespace defines GS-protected namespace the prvHelmClient
+	// is meant for. For App CRs created outside this namespace, the
+	// pubHelmClient should be used
+	privateNamespace = "giantswarm"
```
MarcelMue, line 19: Is this enough? Do we use any other namespaces for our elevated charts? Just checking if there is maybe also kube-system.

service/internal/clientpair/clientpair.go
```diff
@@ -0,0 +1,95 @@
...
+		prvHelmClient: config.PrvHelmClient,
+		pubHelmClient: config.PubHelmClient,
+	}
+
+	return cp, nil
+}
+
+// Get determines which client to use based on the namespace the corresponding App CR
+// is located in. For Workload Cluster, chart operator is permitted to operate under
+// cluster-wide permissions, so there is only prvHelmClient used
```
MarcelMue, line 66:
```suggestion
// cluster-wide permissions, so there is only prvHelmClient used.
```

Turbo nitpick but sentences in godoc should always end with a `.` and they basically never do in this PR :P

service/internal/clientpair/clientpair.go
```diff
@@ -0,0 +1,95 @@
...
+func (cp *ClientPair) Get(ctx context.Context, cr v1alpha1.Chart) helmclient.Interface {
+	// nil pubHelmClient means chart-operator runs in a single-client mode
+	// under cluster admin privileges
+	if cp.pubHelmClient == nil {
+		return cp.prvHelmClient
+	}
+
+	// for App CRs created inside the `giantswarm` namespace, use the prvHelmClient
+	// that runs under cluster admin privileges
+	if key.AppNamespace(cr) == privateNamespace {
```
MarcelMue, line 76: This only guards the annotation. Should we again ensure here that noone was able to add a wrong annotation? E.g. while our webhook was down? So checking additionally `metadata.namespace`

ljakimczuk: The problem is what we get here is the Chart CR, and this is created in the `giantswarm` namespace, the only indication of where the corresponding App CR is located in, is this annotation. I've added a check for it to the app package so it must match the `metadata.namespace` of an App CR when it's set by the user, but no clue what precaution we may use for webhooks being down

MarcelMue: I see - that makes sense to me then :)

MarcelMue (approved):
LGTM in general

service/internal/clientpair/clientpair.go
ljakimczuk: Done, no chart cr found outside the `giantswarm` namespace

kubasobon (approved):
Took another look, LGTM with the updates.

service/internal/clientpair/clientpair.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// WC App Operators are the only cases where the elevated prvHelmClient
+	// has to be used outside the `giantswarm` namespace. It is due to these
+	// apps installing reources the `default:automation` does not have access to.
+	// However, existance of the App CR for unique App Operator outside the
+	// `giantswarm` namespace opens a door to scenarios described here:
+	// https://github.com/giantswarm/giantswarm/issues/22100.
+	// The appOperatorChart defines legal prefix for what elevated Helm Client
+	// can install outside the `giantswarm` namespace.
+	appOperatorChart = "https://giantswarm.github.io/control-plane-catalog/app-operator"
```
uvegla, line 29: When testing changes to `app-operator` using the `control-plane-test-catalog`, will this not be an issue?

ljakimczuk: Depends, if we have any sort of automation that creates `app-operators` from a test catalog in WC namespaces - then yes, I think it may be a problem. I mean, this particular check is there because we have `cluster-apps-operator` creating something outside the `giantswarm` namespace, namely `app-operators` in WC namespaces, that should be reconciled by a privileged (private) Helm client, because these are installed in-cluster. Normally, outside the `giantswarm` we are going to use limited (public) client, but `app-operators` are exception to this rule we needed to account for. I don't know of any controller using the `test` catalog, so I haven't included it, but I'm not 100% sure. Anyway I would be surprised to find such controller. And for us, it doesn't pose a problem because if we need to test new `app-operator` from a `test` catalog, we can still create it, but from the `giantswarm` namespace, instead of WC namespace. I mean, the App CR would look the same, but we would change `metadata.namespace` to `giantswarm` instead of for example `demo0`.

uvegla: Oh ok, so it is `cluster-apps-operator` doing that. I was thinking when using `opsctl deploy` or stopping Flux and maybe manually updating and App CR to test a change it is till `chart-operator` that will pick up the Chart CR for it and deploy the change. In that case the NS is the WC name and the referenced App will be in the test catalog.

ljakimczuk: It's still `chart-operator` who install the Helm Chart, but the App CR is created by `cluster-apps-operator`. In the scanrio you described,  we would need to include `control-plane-test-catalog` because you will be updating App CR that's outside the `giantswarm` namespace. It's like this:

`cluster-apps-operator` --- creates ---> `app-operator` App CR <--- watches for --- unique `app-operator` --- creates ---> `app-operator` Chart CR <---  watches for --- `chart-operator` --- installs ---> Helm Chart

Now, if there is no automation creating in-cluster App CR with `control-plane-test-catalog` we do not need to include it. But if you want to edit App CR that was created by the `cluster-apps-operator`, to change version and catalog, then we need to add the extra catalog to the set.

Another solution for testing I was referring to is about first scaling down the original `app-operator` deployment created for the WC, and then installing new one for testing with App CR created in the `giantswarm` namespace, targetting the WC namespace with `.spec.namespace` of course.

PR #883, opened 2022-06-01
Better support for barebone clusters

whites11:
`chart-operator` is one of the first things to be installed in a management cluster, alongside app-operator.

Back in the day, we used to install cluster-critical components (such as the CNI and kube-proxy) as barebone kubernetes manifests, because the chicken-egg problem prevented us to install apps providing such core components using the app platform.

One day we added the `chartOperator.cni.install` flag that allows to run chart operator even on a cluster lacking core components (specifically the CNI) and thus being able to install the CNI as part of a giant swarm Chart.

In this scenario, `kube-proxy` still needed to be installed as part of the master node bootstrap process which is unfortunate.

This PR makes changes to allow running chart operator even on clusters lacking kube-proxy so we can install kube-proxy as a managed app as well.
This is particularly useful now that we're planning to use `cilium` as a kube-proxy implementation.

## Checklist

### DIFF
--- a/helm/chart-operator/templates/deployment.yaml
+++ b/helm/chart-operator/templates/deployment.yaml
@@ -46,9 +46,6 @@ spec:
       tolerations:
       - effect: NoSchedule
         operator: Exists
-      env:
-        - name: KUBERNETES_SERVICE_HOST
-          value: 127.0.0.1
       {{- else }}
       tolerations:
       - key: node-role.kubernetes.io/master
@@ -84,8 +81,12 @@ spec:
       {{ end }}
       containers:
       - name: {{ .Chart.Name }}
-        {{- if .Values.proxy.enabled }}
+        {{- if or .Values.proxy.enabled .Values.chartOperator.cni.install }}
         env:
+        {{- if .Values.chartOperator.cni.install }}
+        - name: KUBERNETES_SERVICE_HOST
+          value: 127.0.0.1
+        {{- end }}
         {{- if .Values.proxy.http }}
         - name: HTTP_PROXY
           value: {{ .Values.proxy.http }}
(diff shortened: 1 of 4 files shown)


AverageMarcus (approved):
LGTM.

I'm assuming this wont be deployed onto our vintage AWS clusters, right? (If so we might need to think about pods per node again)

helm/chart-operator/templates/deployment.yaml
```diff
@@ -41,11 +41,11 @@ spec:
       serviceAccountName: {{ tpl .Values.resource.default.name  . }}
       {{- if .Values.chartOperator.cni.install }}
       hostNetwork: true
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
```
MarcelMue, line 45: Not sure I agree with this change. The label might change with the rename to `control-plane` and running on control-planes caused trouble before IIRC.

whites11: there is no way around this.
This operating mode is meant to be run during a cluster bootstrap, when CNI and kube-proxy might not be running.
The only way for chart operator to talk to the API server is through 127.0.01 when running inside a master node.

Also, the goal should be to make chart-operator switch to default mode as soon as flux takes over the chart-operator App.

whites11: Actually, no.
In CAPI world it's staying in the `cni.install = true` mode all the time.

See https://github.com/giantswarm/config/blob/main/default/apps/chart-operator/configmap-values.yaml.template#L14 and https://github.com/giantswarm/config/blob/main/installations/gohan/apps/chart-operator/configmap-values.yaml.patch#L3

I think this is very wrong and should be fixed, but that's not this PR's responsibility IMHO

fiunchinho (approved):

PR #884, opened 2022-06-06
Always create `giantswarm-critical` priority class if it does not exist.

whites11:
Chart operator needs a priority class named `giantswarm-critical` to exist.
In MCs and in legacy WCs, we used to create such priority class while deploying master nodes (using k8s-addons script inside master nodes).

In CAPI this does not happen any more, so we needed an alternative way of creating the priority class. For that reason we added a feature flag to `chart-operator`'s helm chart named `cni.install` that, among other things, created the priority class while installing the chart.

This has a bunch of problems:

- makes installing chart operator fail when run with `cni.install = true` on a legacy MC or WC (because the priority class exists already there)
- makes installing chart operator  fail in CAPI clusters when run with `cni.install = false` because the priority class is not present and not created.

To try and solve both issues, this PR simply makes the helm chart create the priority class in any case (regardless of the `cni.install` flag's value) when it does not exist.
This will make the chart work in both the aforementioned scenarios.

WDYT?

## Checklist

### DIFF
--- a/helm/chart-operator/templates/priorityclass.yaml
+++ b/helm/chart-operator/templates/priorityclass.yaml
@@ -1,4 +1,4 @@
-{{- $existing := lookup "scheduling.k8s.io/v1" "PriorityClass" .Release.Namespace "giantswarm-critical" -}}
+{{- $existing := lookup "scheduling.k8s.io/v1" "PriorityClass" "" "giantswarm-critical" -}}
 {{- if not $existing -}}
 apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
--- a/helm/chart-operator/templates/priorityclass.yaml
+++ b/helm/chart-operator/templates/priorityclass.yaml
@@ -1,4 +1,5 @@
-{{- if or .Values.chartOperator.cni.install .Values.e2e  }}
+{{- $existing := lookup "scheduling.k8s.io/v1" "PriorityClass" .Release.Namespace "giantswarm-critical" -}}
+{{- if not $existing -}}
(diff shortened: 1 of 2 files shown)


fiunchinho (approved):
Makes sense

kubasobon (approved):
LGTM

helm/chart-operator/templates/priorityclass.yaml
```diff
@@ -1,4 +1,5 @@
-{{- if or .Values.chartOperator.cni.install .Values.e2e  }}
+{{- $existing := lookup "scheduling.k8s.io/v1" "PriorityClass" .Release.Namespace "giantswarm-critical" -}}
```
kubasobon, line 1: Just one remark: PriorityClasses are cluster-scoped.
```
~ k api-resources | grep -E 'PriorityClass|NAME'
NAME                               SHORTNAMES                 APIVERSION                                 NAMESPACED   KIND
priorityclasses                    pc                         scheduling.k8s.io/v1                       false        PriorityClass
```

whites11: yes but I don't know how it works in the `lookup` function

fiunchinho: I think [you could pass an empty string as namespace](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function)

whites11: ok thanks, changed to empty string

PR #885, opened 2022-06-06
Initial bootstrap mode

whites11:
chart-operator is one of the first things to be installed in a management cluster, alongside app-operator.

Back in the day, we used to install cluster-critical components (such as the CNI and kube-proxy) as barebone kubernetes manifests, because the chicken-egg problem prevented us to install apps providing such core components using the app platform.

One day we added the chartOperator.cni.install flag that allows to run chart operator even on a cluster lacking core components (specifically the CNI) and thus being able to install the CNI as part of a giant swarm Chart.

In this scenario, kube-proxy still needed to be installed as part of the master node bootstrap process which is unfortunate.

This PR adds a new flag `initialBootstrapMode` that relaxes most of chart-operator needs in terms of cluster resources in order to make it run on clusters lacking core components.
This is a backward compatible flag disabled by default, meant to be set to false once the cluster bootstrap is finished.

This is the same behaviour as app-operator

## Checklist

### DIFF
--- a/helm/chart-operator/Chart.yaml
+++ b/helm/chart-operator/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: chart-operator
-version: 1.0.0
-appVersion: 1.0.0
+version: [[ .Version ]]
+appVersion: [[ .AppVersion ]]
 home: https://github.com/giantswarm/chart-operator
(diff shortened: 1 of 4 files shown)


ljakimczuk (approved):

ljakimczuk:
LGTM

PR #897, opened 2022-06-10
Fix missing status on schema validation errors

uvegla:
## Checklist


uvegla:
So this was actually fixed and works since https://github.com/giantswarm/chart-operator/releases/tag/v2.21.1 The latest release still uses https://github.com/giantswarm/chart-operator/releases/tag/v2.21.0, upgrading them will solve the issue.

PR #916, opened 2022-07-20
Use `127.0.0.1` as KUBERNETES_SERVICE_HOST when `bootstrapMode` is enabled

whites11:
When bootstrapMode is enabled, the chart operator pod:

- runs on master nodes only
- runs on hostNetwork

this means we can safely use 127.0.0.1 as the API server address because it's a safe assumption.

This will make bootstrapMode work on workload clusters as well (the current approach of using the node's IP does not work right now because of how the TLS certificates are generated for WCs).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


AverageMarcus (approved):
LGTM

njuettner (approved):

ljakimczuk (approved):

PR #935, opened 2022-08-11
Fix autoscaler

mproffitt:
## VPA Autoscaler fix

Towards https://github.com/giantswarm/giantswarm/issues/23052

This PR introduces changes to the VPA autoscaler configuration that enable pods to be recreated with correct resources in the event the pod is previously OOM killed by the platform. 

If the autoscaler is disabled in the values file or the cluster doesn't support autoscaling, limits are set from the values file, otherwise the autoscaler is enabled with a maximum ceiling of `1000mi` memory and `1000m` cpu.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


kubasobon (approved):
LGTM

PR #953, opened 2022-10-05
Two step installation

ljakimczuk:
## Description

Towards:
- https://github.com/giantswarm/giantswarm/issues/23784
- https://github.com/giantswarm/flux-app/pull/141

## Checklist

### DIFF
--- a/.nancy-ignore
+++ b/.nancy-ignore
@@ -14,6 +14,8 @@ CVE-2022-29153 until=2022-12-30
 CVE-2020-28483 until=2022-12-30
 CVE-2022-24778 until=2022-12-30
 CVE-2022-23648 until=2022-12-30
+CVE-2021-41803 until=2022-12-30
+
 
 # pkg:golang/github.com/nats-io/jwt@v0.3.2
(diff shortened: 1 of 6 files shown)


uvegla:
There is a Helm update called [here](https://github.com/giantswarm/chart-operator/blob/f75f93e3525c6264119d0a087ce09d83d2418baf/service/controller/chart/resource/release/update.go#L114) too. Does that need an update as well?

ljakimczuk:
> Does that need an update as well? 

Nope, the update there runs on different stage of app processing, namely when app is already installed, so we do not need to tweak it in any way.

uvegla:
> > Does that need an update as well?
> 
> Nope, the update there runs on different stage of app processing, namely when app is already installed, so we do not need to tweak it in any way.

When you upgrade an already existing installation it is a helm upgrade, no? In theory it could contain new CRs and resources using them. But I am not sure that is the part of the code that is affected and it is did not come up for now as of yet.

uvegla (approved):

uvegla (approved):

kubasobon (approved):

PR #961, opened 2022-11-02
combine proxy vars from cluster-apps-operator

bavarianbidi:
## Checklist

Towards https://github.com/giantswarm/roadmap/issues/1424


helm/chart-operator/templates/deployment.yaml
```diff
@@ -99,25 +100,27 @@ spec:
...
+        {{- if or (and $proxy.http $proxy.https $proxy.noProxy ) .Values.bootstrapMode.enabled }}
         env:
         {{- if .Values.bootstrapMode.enabled }}
         - name: KUBERNETES_SERVICE_HOST
           value: 127.0.0.1
         - name: KUBERNETES_SERVICE_PORT
           value: {{ .Values.bootstrapMode.apiServerPodPort | quote }}
         {{- end }}
-        {{- if .Values.proxy.http }}
+        {{- if and $proxy.noProxy $proxy.http $proxy.https }}
```
kubasobon, line 111: Will this change the behavior then? To where all of the `$proxy.*` variables have to be set to enable proxy at all?

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+## Added
+
+- Support for running behind a proxy.
+  - `HTTP_PROXY`,`HTTPS_PROXY` and `NO_PROXY` are set as environment variables in `deployment/chart-operator` if defined in `values.yaml`.
```
kubasobon, line 13:
```suggestion
- `HTTP_PROXY`,`HTTPS_PROXY` and `NO_PROXY` are set as environment variables in `deployment/chart-operator` if defined in `values.yaml`.
```

kubasobon (approved):

helm/chart-operator/templates/deployment.yaml
uvegla: I think `cluster-apps-operator` here: https://github.com/giantswarm/cluster-apps-operator/pull/290

bavarianbidi: The behavior itself wont change (beside the fact, that the proxy-vars are set in the `pod.spec` if the vars are set - and not if `proxy.enabled` equals `true`).

It's still possible to define proxy vars on the chart/app. But it's also possible to use the generated proxy variables from `cluster-apps-operator`.

If both proxy values (`cluster.proxy` and `proxy`) are set, the `proxy` values take precedence.

CHANGELOG.md
bavarianbidi: i've set this intentionally but i will adjust the changelog to add some more details about precedence of `proxy` vs. `cluster.proxy`

helm/chart-operator/templates/deployment.yaml
uvegla: What is the case for non CAPx clusters? This is a breaking change for them, right? I mean the values has to be aligned / where do these settings come from on legacy clusters?

bavarianbidi: @uvegla i wasn't aware that we have legacy clusters behind a corporate proxy. I've thought this is a new feature which is only implemented in new CAPI clusters.

The implementation will behave as follows (due some local testing):

* if `cluster.proxy` is defined, these values will be used
* if `proxy` is set and `cluster.proxy` is set, the configured `proxy` values take precedence
* if `proxy` is set and no `cluster.proxy` is set, the `proxy` values will be used.

So i think this isn't a breaking change on vintage clusters (but maybe i'm wrong?)

bavarianbidi: > Will this change the behavior then? To where all of the `$proxy.*` variables have to be set to enable proxy at all?

@kubasobon yes, you're right. Do you think this is a problem at all?

kubasobon: > > Will this change the behavior then? To where all of the `$proxy.*` variables have to be set to enable proxy at all?
> 
> @kubasobon yes, you're right. Do you think this is a problem at all?

I was not sure, but based on the answers here, I think we're safe :)

erkanerol (approved):
LGTM

CHANGELOG.md
jkremser, line 13:
```suggestion
  - `HTTP_PROXY`,`HTTPS_PROXY`, `NO_PROXY` and their lower-case forms are set as environment variables in `deployment/chart-operator` if defined in `values.yaml`.
```

PR #965, opened 2022-11-10
dns-setting-for-private-clusters

calvix:
- Add support to run in  private cloud clusters,  which cannot provide any working `externaLDNSIP`.

when its set to empty value, the dnsPolicy is set to `default` which will pick node DNS and skip the init container part

used with https://github.com/giantswarm/cluster-apps-operator/pull/298


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
+- Add support to run in  private cloud clusters,  which cannot provide any working `externaLDNSIP`.
```
AverageMarcus, line 12:
```suggestion
- Add support to run in private cloud clusters, which cannot provide any working `externalDNSIP`.
```

AverageMarcus (approved):

uvegla (approved):

PR #978, opened 2022-12-14
Add new status for schema validation

ljakimczuk:
## Description

Towards: https://github.com/giantswarm/giantswarm/issues/25052

In short, it is a simple check for the values schema validation failures, that allows us to drop the `unknown-error` status in favour of the `values-schema-violation` that feels more descriptive to the user.

In addition I bumped the `operatorkit` version that solved one CVE, and after merging [this](https://github.com/giantswarm/helmclient/pull/438), I intend to bump the `helmclient` that should solve another one.

## Checklist


uvegla (approved):

service/controller/chart/resource/status/create.go
```diff
@@ -88,7 +88,8 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			} else if cc.Status.Reason != "" {
 				// It can be that there was previous successful deployment and that is the current status of the Helm release
 				// but a follow-up deployment failed for another version, so we want to set a reason indicating the issue
-				reason = fmt.Sprintf("Deployment failed for %#q with %#q: %#q", cr.Spec.Version, cc.Status.Release.Status, cc.Status.Reason)
+				reason = cc.Status.Reason
+				status = cc.Status.Release.Status
```
ljakimczuk, line 92: @uvegla can you take a look at this and tell me if you agree with this change? 

The reason for this change is, because the app, in general, is deployed I understand why the status hasn't been previously changed here, but without using `kgs` (that shows reason), or without `k describe app` (that shows reason as well), there is no way to figure out a problem on the upgrade procedure. Hence, the app may be perceived as `deployed`, where actually it is not, its previous revision is. And it cannot be bumped to the new revision because the error, so to me it feel more semantically correct to change the status, to indicate the problem more clearly.

uvegla: Yea, definitely. We are logging the message and the status but keep the previous one. It is a nice catch and improvement. Just thinking why I feel we already got alerted for this: because the deployed version was different from `spec.version` or something. This will make it nicer and faster to figure out what is up and also for customers who cannot check the app operator logs. 👍🏻

uvegla (approved):

PR #1101, opened 2024-02-08
Log with level error when `helm release` fails

marians:
Little improvement to raise the visibility of actual errors in logs

## Checklist


marians:
Seems like I misunderstood the `Errorf` function. I thought it would mean to emit a message with level error, but it requires the parameter to actually be an error.

```
cannot use "helm release %#q failed, %s" (constant of type string) as error value in argument to r.logger.Errorf: string does not implement error (missing method Error)
```

ljakimczuk:
@marians you can pass to it the `err` you have access to

PR #1167, opened 2024-07-17
Enable an emptyDir volume under /tmp, make root fs read only and thus align resources and correct VPA usage

uvegla:
Towards: https://github.com/giantswarm/giantswarm/issues/31166

Please note the VPA related changes and the changes to default resource requests and limits because the emptyDir volume will contribute to the memory usage.

## Checklist


uvegla:
✅ Tested WC creation with this version.

uvegla:
@stone-z (cc. @teemow) This seems to work, in reaction to https://github.com/giantswarm/giantswarm/issues/31166#issuecomment-2208805758 and also thus sets `readOnlyRootFilesystem: true` on the security contexts. This of course can be changed by values, but was set as the default behaviour. Does that satisfy the original claims in https://github.com/giantswarm/giantswarm/issues/31160?

helm/chart-operator/values.yaml
```diff
@@ -18,9 +18,22 @@ clusterDNSIP: 172.31.0.10
...
+    memory: 256Mi
   limits:
-    memory: 350Mi
+    cpu: 100m
+    memory: 512Mi
+
+verticalPodAutoscaler:
+  enabled: true
+  maxAllowed:
+    cpu: 150m
```
stone-z, line 29: Is 150m enough on big clusters?

stone-z (approved):
This is an overall improvement even aside from the security impact :+1:

helm/chart-operator/values.yaml
uvegla: Hmm, the previous max allowed was hard coded on the VPA to be 1 full core. Maybe we can increase it a little further, but I imagine the bottleneck on the memory.

uvegla: Yea, good point. I looked at some aggregated data about CPU requests across all our clusters. Let me adjust it.

uvegla: Updated as of: https://github.com/giantswarm/chart-operator/pull/1167/commits/3f3835f095d1afa8a4931006acd79cef37cf30a6

stone-z (approved):
Looks good

PR #1178, opened 2024-10-18
Fix the `pending-upgrade` problem

ljakimczuk:
Towards: https://github.com/giantswarm/giantswarm/issues/29793, https://github.com/giantswarm/giantswarm/issues/31980

Chart Operator on preparing update checks if Helm release remains in the `pending-*` status longer than pre-configured timeout for the operation, and if so, it tries to recover from it by changing the status of a release to `failed` or `unknown`, depending on a revision of this release.

## Checklist


mproffitt (approved):
Overall this looks fine to me. I'm satisfied that the approach is suitable for the current state of the operator

